PR #26, opened 2023-08-02
Review and implement new test format

yangsophieee:
Reviewed existing tests (have since replaced the old test-xamples.R with new test format). The new test format is now comprised of folders in tests/testthat/examples, called Test_2023_1, Test_2023_2, etc. containing data.csv, metadata.yml, README.md and an output folder. The output folder is generated by building the datasets via `test_build_dataset` and then using the `write_plaintext` function to generate the tables; these will eventually be modified so that the output is actually what we expect (right now, some issues exist but we want the tests to pass). The tests in test-xamples.R now test whether `test_build_dataset` generates the same output as the tables in the expected output folder. 

Right now Test_2023_1 and Test_2023_2 are modified versions of Falster_2005_1. The rest of the test datasets have not yet been modified, but the README files are half completed. 

Some issues with this new test format currently exist:
- When you run the `test_that` chunk in test-xamples.R, it generates a different output to running each line within the chunk one-by-one. The difference is the id's are all generated differently (e.g. `entity_context_id,` `method_id,` etc.). I've fixed it for now by generating the expected output via a `test_that` chunk. This is really weird.
- Writing the test dataset output to plain text then reading the output back in for testing generates differences with column data types, `names` of columns and so on. I've changed the expected output manually within the test_that chunk to get them to be the same for now. See "# Temporary modifications to get these tests to pass". 
- I've also fixed some tidyselect warnings after running `test_check` on the package.

I'm very sorry for all the formatting changes, will not do the same next time :(


yangsophieee:
@dfalster I've added the changes we've discussed!

dfalster (approved):
Great work here Sophie. This new structure will be very useful and you've implemented it very nicely. Thank you.

Please merge with a squash commit

Issue #3, opened 2023-06-28
Duplicate context values do not get collapsed


yangsophieee:
Referencing this [pull request](https://github.com/traitecoevo/austraits.build/pull/749#issue-1776025544) from [austraits.build](https://github.com/traitecoevo/austraits.build):

Currently if you have multiple context values with the same `value:` field (but maybe different `find:` fields) AusTraits will not collapse those duplicate values when building the `austraits$contexts` table. When you pivot the context table wider, as there are non-unique rows, this results in a wide table that has list data types where there should be character data types. This was an issue for Firn_2019, Hall_1981, vanderMoezel_1987 and Wright_2001.

I've added `custom_R_code` to remove the duplicate context values from the context metadata but for the future we will try to fix the AusTraits building process.

yangsophieee:
For Lines [470](https://github.com/traitecoevo/traits.build/blob/d1ed8fc1ec6e20528f59a2b450244457119ba81a/R/process.R#L470)-477, the code doesn't line up with what the comment says the code should do. `find` values are not replaced by `value`'s. I assume the goal is to replace `find` values with `value` values so that the id's can generate correctly. [Line 472](https://github.com/traitecoevo/traits.build/blob/d1ed8fc1ec6e20528f59a2b450244457119ba81a/R/process.R#L472)-473 filters for rows that have NA values, instead of non-NA values. The second argument on [Line 475](https://github.com/traitecoevo/traits.build/blob/d1ed8fc1ec6e20528f59a2b450244457119ba81a/R/process.R#L475) should probably be `xx$find`.

However Line 472-473 might be redundant anyway. On Line 443, NA `find` values are replaced with the `value` values, if a `find` column already exists in the context metadata. This means Line 472 will never find NA values in these circumstances. Line 545 is also redundant because of Line 443.

Also, I think Line 471 is unnecessary because the find column is always present because of [Line 441](https://github.com/traitecoevo/traits.build/blob/d1ed8fc1ec6e20528f59a2b450244457119ba81a/R/process.R#L441) in [process_format_contexts](https://github.com/traitecoevo/traits.build/blob/d1ed8fc1ec6e20528f59a2b450244457119ba81a/R/process.R#L421C1-L421C24).

[...]

Issue #2, opened 2023-06-28
Some numeric contexts are not being read in


yangsophieee:
Referencing this [issue](https://github.com/traitecoevo/austraits.build/issues/745) and [comment](https://github.com/traitecoevo/austraits.build/pull/749#issuecomment-1610297632), brought to our attention by @ehwenk:
 
Some numeric context values were not being read in, e.g. Canham_2023 and Witkowski_1991. These will be instances where context provides background information for a trait value, but is not required to uniquely identify rows of data. A dataset test to create for this would be, for each category, to determine how many distinct context value combinations there are within a dataset for each context category and confirm that is equal to the number of id's generated (i.e. for each type of context identifier in `austraits$traits`).

@ehwenk also noticed that if you change the context columns to `as.character()` in the `custom_R_code` it solves the issue. This seems like something that can be easily fixed in `process.R`.

yangsophieee:
I think I may have figured out why the numeric contexts don't get read in. On [Line 477](https://github.com/traitecoevo/traits.build/blob/d1ed8fc1ec6e20528f59a2b450244457119ba81a/R/process.R#L477C52-L477C52) of process.R it uses the named vector on Line 476 to replace the `context_cols` values. Since the value types in the named vector are character types it requires the data type in `context_coles` to also be character type, otherwise you get no matches (leading to a bunch of NAs). So that's why changing the context columns to character type in `custom_R_code` fixes this issue.

I also noticed though that the code doesn't line up with what the comments say the code should do. I played around with the code and committed it on [this branch](https://github.com/traitecoevo/austraits.build/tree/fix-contexts-not-being-used) but reverted it since I think the issue might be a bit larger than just those few lines. Seeing the overall workflow I think I found the reason why [duplicate context values are not getting collapsed as well](https://github.com/traitecoevo/traits.build/issues/3) and will detail it on the other issue.

PR #37, opened 2023-08-03
Collapse duplicate contexts and fix contexts not being used

yangsophieee:
Fixes these two issues: https://github.com/traitecoevo/traits.build/issues/3 and https://github.com/traitecoevo/traits.build/issues/2.

The `contexts` table should now match the expected output `contexts` table, with no duplicate context values (https://github.com/traitecoevo/traits.build/issues/3). 

This should also fix the issue with contexts not being 'read in' (https://github.com/traitecoevo/traits.build/issues/2), as was happening with Canham_2023 and Witkowski_1991 in AusTraits. This was fixed by converting the context columns to character type ([Line 490](https://github.com/traitecoevo/traits.build/blob/f8e4a36cc5df6e0feca41f84d3c23609e8c613f9/R/process.R#L490)) because replacing `find` values with `value` values on Line 504 and [Line 506](https://github.com/traitecoevo/traits.build/blob/f8e4a36cc5df6e0feca41f84d3c23609e8c613f9/R/process.R#L506) can only work if they are character types.


R/process.R
```diff
@@ -548,12 +552,12 @@ process_create_context_ids <- function(data, contexts) {
     for (v in vars) {
       id_link[[v]] <-
         xxx %>%
-        dplyr::rename(dplyr::all_of(c(find = v))) %>%
-        dplyr::select(dplyr::all_of(c("find", "id"))) %>%
+        dplyr::rename(dplyr::all_of(c(value = v))) %>%
```
dfalster, line 555: I can't recall enough of the logic here. Can you elaborate on why we should be suing value instead of find?

R/process.R
```diff
@@ -1166,6 +1169,7 @@ process_parse_data <- function(data, dataset_id, metadata, contexts) {
   }
 
   # Implement any value changes as per substitutions
+  # I think this breaks if substitutions is .na in the metadata?
   if (!is.na(metadata[["substitutions"]][1])) {
```
dfalster, line 1173: Please check if it breaks. the format `is.na(metadata[["substitutions"]][1]`is intended to catch instances where `substitutions: .na` in the yaml

tests/testthat/examples/Test_2023_1/metadata.yml
```diff
@@ -71,8 +71,7 @@ contexts:
   values:
   - value: female
     description: female plant
-  - find: male
-    value: male
+  - value: male
```
dfalster, line 74: Agree this one doesn't make sense. Please ensure there are examples in the tests that do have a reevant example using `find`

dfalster (changes requested):
Nice work! Some queries to be resolved

tests/testthat/examples/Test_2023_1/metadata.yml
yangsophieee: Yes in the same 'plant sex' `context_property` I put ['female2'](https://github.com/traitecoevo/traits.build/blob/eb4a62ed0efe066744e71af95eb2f4fa172c88ef/tests/testthat/examples/Test_2023_1/metadata.yml#L76) in `find` and 'female' in `value`.

R/process.R
yangsophieee: On [Line 501](https://github.com/traitecoevo/traits.build/blob/eb4a62ed0efe066744e71af95eb2f4fa172c88ef/R/process.R#L501), I filtered for rows where the `find` column is not NA. If there are rows where the `find` column is not NA, I create a named vector with the `value` column as the values and `find` column as the names. Then on [Line 506](https://github.com/traitecoevo/traits.build/blob/eb4a62ed0efe066744e71af95eb2f4fa172c88ef/R/process.R#L506), it uses this named vector to replace the `find` context values with `value` values. We are now working completely with `value` values and can ignore the `find` column. 

That's why on [Line 554](https://github.com/traitecoevo/traits.build/blob/eb4a62ed0efe066744e71af95eb2f4fa172c88ef/R/process.R#L554) it's now creating link values via the `value` column. We don't want to create link values via the `find` column because the `find` column doesn't contain the final unique context values. 

Let me know if it doesn't make sense!

R/process.R
yangsophieee: All good, it does not break. It was only breaking for Canham_2023 because `read_yaml` wasn't able to read a special character (Ψ) so it was never able to read substitutions (hence substitutions was NULL instead of NA).

yangsophieee:
@dfalster Added a few minor changes (mostly comments). Let me know if it's ready to squash and merge :)

dfalster (approved):
Nice work!

Issue #39, opened 2023-08-07
metadata_add_locations not working


ehwenk:
The function metadata_add_locations is not working when sourced from traits.build

```
metadata_add_locations("Crous_2013", locations)
Select column for `location_name` 

1: Location
2: longitude (deg)
3: latitude (deg)

Selection: 1
Error in dplyr::all_of(c("location_data"), -!!location_name) : 
  unused argument (-!!location_name)
```

The function works fine when sourced from austraits.build:

```
library("austraits.build")

metadata_add_locations("Crous_2013", locations)
Select column for `location_name` 

1: Location
2: longitude (deg)
3: latitude (deg)

Selection: 1
Indicate all columns you wish to keep as distinct location_properties in  Crous_2013 (by number separated by space; e.g. '1 2 4'):
1: longitude (deg)
2: latitude (deg)

Selection: 1 2
Following locations added to metadata for Crous_2013: Richmond NSW, Australia
	with variables longitude (deg), latitude (deg).
	Please complete information in data/Crous_2013/metadata.yml.
```

PR #40, opened 2023-08-09
Fix `metadata_add_location`

yangsophieee:
Arising from issue https://github.com/traitecoevo/traits.build/issues/39:
Issue probably arose from adding in `tidyselect` terms `all_of` and `any_of` a while back. I'm assuming during this process that the first argument of select (`.data`) was erroneously put into an `all_of` argument. 

I did some research on the syntax `!!` and I'm not sure if it's necessary here @dfalster. I can't think of any instances where simply using `-dplyr::all_of(c(location_name))` wouldn't work but please correct me if I'm wrong!

### DIFF
--- a/R/setup.R
+++ b/R/setup.R
@@ -276,8 +276,11 @@ metadata_add_locations <- function(dataset_id, location_data) {
 
   cat(
     sprintf("Following locations added to metadata for %s: %s\n\twith variables %s.\n\tPlease complete information in %s.\n\n",
-    dataset_id, crayon::red(paste(names(metadata$locations), collapse = ", ")),
-    crayon::red(paste(keep, collapse = ", ")), dataset_id %>% metadata_path_dataset_id())
+      dataset_id,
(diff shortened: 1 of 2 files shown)


dfalster (commented):
Hi @yangsophieee 

Nice. I've suggested a further simplifications with my recent commits. 

The biggest test is whether it works, I've not tested this. If you can confirm then happy to merge

yangsophieee:
Hi @dfalster, I've tested it on Crous_2013, it works now. I've added a test in test-setup.R to catch errors. 

I also noticed that `metadata_add_locations` replaces any existing location metadata. Do you think it would be good if it asked the user if they want to replace existing location metadata (or maybe append it)? @ehwenk

yangsophieee:
Oops, seeing your issue https://github.com/traitecoevo/traits.build/issues/41, I've removed the test I made as I'm not sure if it works as intended. I'll have a look at proper testing.

dfalster (approved):

Issue #41, opened 2023-08-09
Test functions requiring user input


dfalster:
We currently don't tests any of the functions requiring interactive user input. This let some errors slip in (see #39 ). We should be able to adjust the interface slightly to enable testing. For example, see these posts

- https://stackoverflow.com/questions/65740390/how-to-test-a-function-that-depends-on-a-menu-user-input
- https://debruine.github.io/post/interactive-test/
- https://stackoverflow.com/questions/41372146/test-interaction-with-users-in-r-package

Ideally, these three functions would get tests added at some point

- metadata_add_locations
- metadata_add_contexts
- metadata_create_template

yangsophieee:
`metadata_add_traits` also needs tests I think.

PR #42, opened 2023-08-09
Add tests for functions containing `utils::menu`

yangsophieee:
Relating to this issue (https://github.com/traitecoevo/traits.build/issues/41):
I've modified `metadata_add_locations` and implemented a test structure in `test-setup.R` for this function, based on [this thread](https://stackoverflow.com/a/65935322). 

I added an ifelse statement to `metadata_add_locations`, where the logic is that if `metadata_add_locations` is NOT run inside a `test_that` chunk, user input will be requested via `utils::menu` as normal. If it's run inside a `test_that` chunk then the function will simply return `names(location_data)`. Then in test-setup.R it tests if `metadata_add_locations` returns `names(location_data)`. I'm not sure what else to test, because right now it seems a bit pointless to test whether `names(location_data)` is what we expect. It doesn't test whether all the user input leads to the correct outcome. 

Within the else statement you could manually prescribe the first and second column of `location_data` to `location_name` and `keep` without user input and test the rest of the function that way. This feels clunky. 

It also feels wrong to modify the body of `metadata_add_locations` specifically for accommodating tests. 

I tried using the `mockr` package as suggested [here](https://stackoverflow.com/a/65740756) but couldn't get it to work. Another option is to replace `utils::menu` with `readLines` which seems quite promising (https://debruine.github.io/post/interactive-test/). Would you be happy with replacing `menu` with `readLines`?

Let me know what you think is the best option and I'll add similar modifications to `metadata_add_traits`, `metadata_add_contexts` and `metadata_create_template`.


R/setup.R
```diff
@@ -284,6 +291,8 @@ metadata_add_locations <- function(dataset_id, location_data) {
   )
 
   write_metadata_dataset(metadata, dataset_id)
+
+  return(invisible(metadata))
```
dfalster, line 295: even though we saved to file, this will return the object silently, which enables us to test it

tests/testthat/test-setup.R
```diff
@@ -102,6 +102,26 @@ test_that("metadata_add_source_bibtex is working", {
...
+  # Hence only works when run within the chunk, not line-by-line
+  expect_no_error(
+    suppressMessages(
+    x <- metadata_add_locations("Test_2022", locations, 
+      # gives responses for user input, for testing
+       user_responses = list(location_name = "site_name", keep = c("latitude", "longitude", "elevation"))
+      )
+    )
+  )
+  expect_equal(names(x$locations) , locations$site_name)
```
dfalster, line 122: Now we can actually tests if the object was modified

tests/testthat/functions.R
```diff
@@ -142,3 +142,16 @@ test_structure <- function(
   # Contains allowed traits
   expect_isin(data$traits$trait_name %>% unique(), definitions$elements %>% names(), info = paste("traits ", v))
 }
+
+## a helper function to determine if this is being run as part of a test
+is_testing_env <- function() {
```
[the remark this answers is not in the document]
yangsophieee: This function isn't being used in any code though is it?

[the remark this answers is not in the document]
dfalster: No, it was, but then I switched strategy to providing inputs and using that for if statement. I N left it in FYI

yangsophieee:
Hi @dfalster, thank you so much! I completely agree that my testing wasn't actually testing the function but I wasn't sure how best to do it. Your code looks very elegant; thanks for lending your expertise 😄. Did you want me to remove the `is_testing_env` function? Happy to merge this in now.

dfalster (changes requested):
Can you add more tests on output, then replicate this strategy for other functions requiring user input?

yangsophieee:
I've added optional arguments for simulated user inputs for `metadata_add_locations`, `metadata_add_traits`, `metadata_add_contexts` and `metadata_create_template`. I've also revamped the console messages and added more testing for the functions. I also updated documentation of functions (all in the man folder; should I have done this on a separate branch?). 

There were a few functions that had no testing which I've since added tests for: `metadata_find_taxonomic_change`, `build_find_taxon`, `metadata_add_taxonomic_changes_list`, `metadata_add_substitutions_list`. Not sure how useful these last few functions are. The difference between `metadata_add_substitutions_table` and `metadata_add_substitutions_list` isn't large and the naming does not help to elucidate the differences. Maybe we could remove `metadata_add_substitutions_list` and rename `metadata_add_taxonomic_changes_list` to `metadata_add_taxonomic_changes_table`? I'm sure there are a lot of improvements that can still be made that future users will bring to our attention. For now I think it's good to merge and I can move on to fixing issues.

tests/testthat.R
```diff
@@ -1,4 +1,4 @@
-library(testthat)
+library(testthat) # Do you need this line if test_check has `testthat::` in front of it?
```
dfalster, line 1: No, delete

tests/testthat.R
```diff
@@ -1,4 +1,4 @@
...
 library(traits.build)
```
dfalster, line 2: I think we can delete this too

tests/testthat/test-utils.R
```diff
@@ -1,15 +1,15 @@
 
-test_that("util_replace_null returns NA", {
+test_that("`util_replace_null` returns NA", {
   expect_equal(util_replace_null(NULL), NA)
   expect_equal(util_replace_null(1), 1)
 })
 
-test_that("util_df_to_list is working", {
+test_that("`util_df_to_list` is working", {
   expect_type(util_df_to_list(ggplot2::mpg), "list")
```
dfalster, line 8: Nice idea, but we don't want to rely on ggplot package for this, as it's not a core dependency and we should keep the footprint (list of dependencies as small as possible). Can we use one of the dfs in base R or dplyr (which is a core dependency)?

E.g. `iris` or `dplyr::band_instruments` or `dplyr::starters`?

tests/testthat/test-utils.R
```diff
@@ -1,15 +1,15 @@
...
 })
 
 # There's no test for util_list_to_df1, should we add one?
-test_that("util_list_to_df2 is working", {
+test_that("`util_list_to_df2` is working", {
   expect_equal(util_list_to_df2(NULL), NA)
   expect_equal(util_list_to_df2(NA), NA)
```
dfalster, line 15: We should add a couple more tests, .e.g to show that the function produces output in desired format

`traits.build::util_df_to_list(dplyr::band_instruments)`

dfalster (approved):
Nice work @yangsophieee . Very thorough and needed. Before merging, please request a review from Lizzy, ask her to run everything as though she were inputting a new dataset. She doesn't need to review code.

Issue #50, opened 2023-08-30
`metadata_add_contexts` should not write NAs to the context metadata


yangsophieee:
Right now `metadata_add_contexts` propagates NAs to the context metadata (see below):
![image](https://github.com/traitecoevo/traits.build/assets/78716301/bd336c70-0fe5-451d-981e-e22f4a35fabf)
![image](https://github.com/traitecoevo/traits.build/assets/78716301/ea243a40-0fef-41dc-987c-03c0e4466c7b)
This is misleading as NAs should not be in the `find` field, nor in the `value` field (unless there is an accompanying `find` field).

Issue #47, opened 2023-08-30
[traits.build testing] Testing for `dataset_test`


yangsophieee:
Currently there isn't testing for `dataset_test` except a few basic lines checking that it works with argument `reporter = testthat::SilentReporter`. It would be nice if it could test that certain messages exist in the output when there are issues in the metadata.

ehwenk:
solved with solved with https://github.com/traitecoevo/traits.build/commit/0f0517dba4305bf1a582cc2e5f010d3189721de7

Issue #32, opened 2023-08-03
False positive error in `dataset_test`


yangsophieee:
Arising from [this issue](https://github.com/traitecoevo/ausinvertraits.build/issues/68) in AusInverTraits via @janeeogilvie:

"I am getting this error for all of the studies that I am adding the context_property "plant association" to, for which I am specifying replacement values. Note that the value replacing is working, so the error may be incorrect."
![image](https://github.com/traitecoevo/traits.build/assets/78716301/4a0345ff-4f00-4e6e-882f-4a55c30c1c6f)

I investigated the error and it is indeed a false positive. Removing the context metadata for `na.character` will fix it, but we should implement it so that `na.character` is either not allowed to be entered as a context metadata or it can be read in with no issues.

yangsophieee:
I think this issue might still be happening... investigating now.

yangsophieee:
I've fixed the issue on `read_metadata_fields_from_column_at_trait_level` by adding [this line](https://github.com/traitecoevo/traits.build/blob/3ba84a9937956d64fd186c69526fff5f40583fc5/R/process.R#L470). 

```
    if (is.null(contexts[["find"]])) {
      # Add a `find` column
      contexts[["find"]] <- NA_character_
      # Replace NAs with `value` values (does not affect later code but is important for Line 549 in testdata.R
      # so that `find` column isn't all NA)
      contexts[["find"]] <- ifelse(is.na(contexts$find), contexts$value, contexts$find)
```

Issue #22, opened 2023-07-21
Missing `link_id` and `link_vals` for contexts in AusInverTraits


yangsophieee:
Arising from [this AusInverTraits issue](https://github.com/traitecoevo/ausinvertraits.build/issues/69):

When adding two context properties to the metadata files ('associated plant taxa' and 'plant association'), the `link_id` and `link_vals` are missing for most studies (see the attached screenshot).

![image](https://github.com/traitecoevo/traits.build/assets/78716301/0e853b5e-9c96-4835-a511-cb38ea74872e)

Note that this issue is resolved when metadata context for '.na.character' is removed.

PR #51, opened 2023-08-30
Fix ausinvertraits bug

### LINKED ISSUE #35: No `link_id` and `link_vals` generated for "sampling time of day" context in Test_2023_1
There are no `link_id` and `link_vals` generated for one of the "sampling time of day" context values in Test_2023_1. The only context value is being generated correctly. 

image

yangsophieee:
Sorry for the long list of formatting edits... I need to make a separate branch next time. Also I'm addressing a lot of different issues here, sorry 😢  

Here's a summary of all the changes that aren't formatting changes:
- Removed Line 508-509 in process.R (which filters out NA `find` values) (now [Line 498](https://github.com/traitecoevo/traits.build/blob/c00a2191af2eb37255f8ddd7e8b051bcbc3939e1/R/process.R#L498)) because there will never be any NA `find` values since [Line 469](https://github.com/traitecoevo/traits.build/blob/c00a2191af2eb37255f8ddd7e8b051bcbc3939e1/R/process.R#L469) replaces them with the `value` field. 
- Added [Line 566](https://github.com/traitecoevo/traits.build/blob/c00a2191af2eb37255f8ddd7e8b051bcbc3939e1/R/process.R#L566) in process.R because we don't want NAs to be in the contexts table and also don't want NAs in the traits table to be assigned `context_id` when `left_join`ing. 
- I forgot to add this into the other branch that just got merged in, but on [Line 440](https://github.com/traitecoevo/traits.build/blob/c00a2191af2eb37255f8ddd7e8b051bcbc3939e1/R/setup.R#L440) of setup.R I added `na.omit` so that NAs are not added into context metadata. (#50)
- [Line 450](https://github.com/traitecoevo/traits.build/blob/c00a2191af2eb37255f8ddd7e8b051bcbc3939e1/R/testdata.R#L451)-504 in testdata.R: I've added tests to check that `find` values should not be NA, `find` is accompanied by a `value` field, `description` should not be accompanied by an NA `value` field, and `description` is accompanied by `find` and/or `value` fields. This will solve #22 which was occurring because of various NA `find` and `value` values. 
- [Line 539](https://github.com/traitecoevo/traits.build/blob/c00a2191af2eb37255f8ddd7e8b051bcbc3939e1/R/testdata.R#L539) in testdata.R: I'm a bit confused on how context values can align with a column in the traits metadata table ([Line 515](https://github.com/traitecoevo/traits.build/blob/c00a2191af2eb37255f8ddd7e8b051bcbc3939e1/R/testdata.R#L515)). 
- [Line 546](https://github.com/traitecoevo/traits.build/blob/c00a2191af2eb37255f8ddd7e8b051bcbc3939e1/R/testdata.R#L546) in testdata.R: I removed code that was causing a bug for AusInverTraits people (#32). I think it makes sense to just look for context values in the `find` column.
- Fixed some mistakes in Test_2023_1 (Fixes #35).
- Added Test_2023_7 for the hopes of later implementing testing for `dataset_test` (#47).
- Removed cols() argument from Line 590 of vignettes/file_structure.Rmd, otherwise custom R code that manipulates numeric data types won't work on character types.


tests/testthat/export/reports/Test_2022.html
dfalster, line 1: This file shouldn't be tracked or updated

dfalster (approved):
Looks good Sophie

tests/testthat/export/reports/Test_2022.html
yangsophieee: Yeah I was wondering why it was tracked in the first place! I've removed it.

PR #53, opened 2023-08-31
Replace superseded functions `mutate_all`, `separate_rows` and minimise R CMD check warnings

yangsophieee:
Replaced superseded functions `mutate_all` and `separate_rows`, cleaned up use of `rename` and minimised R CMD check warnings (e.g. 'no visible binding for global variable x') (#13).


dfalster (approved):
Great work!

Issue #15, opened 2023-06-30
Add functionality: for context, omit values & description from metadata.yml file


dfalster:
Arising from https://github.com/traitecoevo/austraits.build/issues/673 via @ehwenk 

Currently, each value for a context property and a description for each value are required to be listed in the metadata.yml file in order to be read into austraits. 

This is unnecessary for continuous context properties, where each row of data will have a unique context property and no descriptions are required.

We should have it that all the unique context property values from the specified column are read into austraits even if they are not listed in the metadata.yml file. In these cases, the description will simply remain "unknown".

See Prior_2022 as an example of the very long metadata.yml file that is generated currently

ehwenk:
closed with commit [3e037af](https://github.com/traitecoevo/traits.build/commit/3e037af4cba612e2f774dd702171ce5285036077)

PR #59, opened 2023-09-06
Omits lists of values from contexts

ehwenk:
Rework contexts such that values are no longer required to be listed under context properties.

If neither `descriptions` nor replacement values need to be added to the context property values being read in, then the context field in the metadata[["contexts"]] now will just include the rows for `context_property`, `category`, and `var_in`. The context property values are then sourced during processing as the unique values from the specified `var_in` column. This greatly reduces the length of the metadata.yml file for datasets where there are context_properties that are continuous variable (i.e. plant diameter, leaf temperature)

Solves issue #15


yangsophieee:
I've added a test in Test Dataset 1 (extra context property called "plant diameter" that omits values from context metadata). Seems to be working great!

tests/testthat/examples/Test_2023_1/output/traits.csv
```diff
@@ -1,123 +1,123 @@
 dataset_id,taxon_name,observation_id,trait_name,value,unit,entity_type,value_type,basis_of_value,replicates,basis_of_record,life_stage,population_id,individual_id,temporal_id,source_id,location_id,entity_context_id,plot_id,treatment_id,collection_date,measurement_remarks,method_id,original_name
-Test_2023_1,Acacia celsa,001,flowering_time,nyynnnnnnnnn,,individual,mode,expert_score,3,field,adult,01,01,02,,01,03,01,01,2004/2004,from Falster_2005_1,07,Acacia celsa
```
dfalster, line 2: Can you confirm that all these changes I outputs are what you intended? If so, happy for you to proceed

dfalster (approved):
Looks good. See one query.

tests/testthat/examples/Test_2023_1/output/traits.csv
yangsophieee: It looks good to me because I added a new `entity_context` called "plant diameter" which should have changed all the `entity_context_id` numbers, which it looks like has happened.

R/process.R
```diff
@@ -430,44 +434,79 @@ process_generate_id <- function(x, prefix, sort = FALSE) {
...
+      contexts <-
       my_list %>%
-      purrr::map_df(f) %>%
-      dplyr::mutate(dataset_id = dataset_id) %>%
-      dplyr::select(dplyr::any_of(
-        c("dataset_id", "context_property", "category", "var_in",
-        "find", "value", "description"))
+      purrr::map(f, dataset_id)
+    
+    for (i in 1:length(contexts)) {
```
dfalster, line 467: I'm going to rework this, we may as well put all the stuff below into the function above, as it already works through each context. 

There's no need creating these groups with a map, then looping over them.

R/process.R
```diff
@@ -430,44 +434,79 @@ process_generate_id <- function(x, prefix, sort = FALSE) {
...
+      ## if the field `description` is missing from metadata[["contexts"]] for the specific context property, create a column now 
+      if (!"description" %in% names(contexts[[i]])) {
+        contexts[[i]]$description <- NA_character_
+      }
+      
+      ## if the fields `find` and `value` are both missing from metadata[["contexts"]] for the specific context property create them
+      ## they are both the unique set of values in the column in the data.csv file.
+      if(!"find" %in% names(contexts[[i]]) & 
+         !"value" %in% names(contexts[[i]])) {
+        to_join <- unique(traits[[contexts[[i]]$var_in[1]]]) %>%
```
dfalster, line 478: The `[1]` in this line suggests it will only sue the first instance of `var_in`, is that intended?

dfalster (changes requested):
I can help make suggested changes

dfalster (approved):

Issue #58, opened 2023-09-06
Correctly read `entity_type` and `value_type` as a fixed value when a column of the same name exists


yangsophieee:
Previously, `entity_type` would be read in incorrectly any time the fixed value was the same as a column name in the data (e.g. `species` is the fixed value and there's also a column in the data called `species`, which was a very common occurrence). This may also be an issue with `value_type` if there's columns named 'minimum', 'maximum', 'mean', etc. Lizzy has added in code that Daniel made a while ago addressing this issue on the `solving-minor-issues` branch. Basically it looks for whether any data columns match `entity_type` or `value_type` values in the schema, and if so, it won't read it in as a column.

My question is whether it should look at the schema for all the metadata fields (`basis_of_value`, etc.), not just `entity_type` and `value_type`. It will be unlikely columns will be named after these values but it could happen. 

There was also a minor issue that may be related, where `value_type` got introduced into the `methods` table, but I've since fixed it on the `read_metadata_fields_from_column_at_trait_level` branch by adding it to [Line 1292](https://github.com/traitecoevo/traits.build/blob/5f0a6a08d9e2ab36385f568344431da4bae250eb/R/process.R#L1292) of `process.R`. Have since deleted it from the schema file too.

yangsophieee:
I think I'm going to optimise the fix slightly by putting `not_allowed` in the `for` loop cycling through `vars_to_check`. So that for each variable in `vars_to_check` it will look at the schema for that specific variable's values instead of all of them. 

That is, right now it's:
```
  not_allowed <- c(
    schema[["entity_type"]][["values"]] %>% names(),
    schema[["value_type"]][["values"]] %>% names()
  )
```
But now after Line [1123](https://github.com/traitecoevo/traits.build/blob/ae2e8966e8fc12b48d88bd7d17b1896a2cfc8219/R/process.R#L1123):
```
not_allowed <- schema[[v]][["values"]] %>% names()
```
where `v` is `"entity_type"` and `"value_type"`, etc.

ehwenk:
closed with [8714247](https://github.com/traitecoevo/traits.build/commit/8714247b22744216c23d2c533f277b576b5d5d5f)

PR #64, opened 2023-09-06
Read metadata fields from column at trait level for long datasets

yangsophieee:
Allow reading of metadata fields from column at trait level or dataset level for long datasets (#6)
- Added tests for the above
- Also optimised the fix for #58 (see [this comment](https://github.com/traitecoevo/traits.build/issues/58#issuecomment-1707835927) for my fix)
- Fixed issue #32 which I thought I had fixed but there was an edge case (see [this comment](https://github.com/traitecoevo/traits.build/issues/32#issuecomment-1707890518) for my fix) 
- I removed `value_type` from the methods table and the schema as Lizzy decided they shouldn't be there


yangsophieee:
I'll let Lizzy's pull requests be merged in first as I think some of them may conflict with my fixes.

dfalster (approved):
Well done Sophie

This is a lot of work! 

I'm relying on the new tests to check the intended behaviour.

Issue #65, opened 2023-09-06
Warning during unit conversions


ehwenk:
This ongoing warning when building AusTraits is continuing to baffle us:
```
Warning in View :
  There was 1 warning in `dplyr::mutate()`.
ℹ In argument: `value = ifelse(...)`.
ℹ In group 2: `ucn = "a-mo"`, `to_convert = TRUE`.
Caused by warning in `unit_conversion_functions[[name]]()`:
! NAs introduced by coercion
```
@yangsophieee and I assumed it was because of measurements where `value_type = bin` that required unit conversions. But that is now solved and the error persists. 

These lines of code in process.R cause the problem:
```
 data %>%
    dplyr::group_by(.data$ucn, .data$to_convert) %>%
    dplyr::mutate(
      value = ifelse(.data$to_convert & !.data$value_type %in% c("bin", "range") & !is.na(.data$value), 
                     f(.data$value, .data$ucn[1]), .data$value), #exclude bins and ranges, or get a warning
      split_values_1 = ifelse(.data$to_convert & .data$value_type %in% c("bin", "range") & !is.na(.data$split_values_1), 
                              f(.data$split_values_1, .data$ucn[1]), .data$split_values_1),
      split_values_2 = ifelse(.data$to_convert & .data$value_type %in% c("bin", "range") & !is.na(.data$split_values_2), 
                              f(.data$split_values_2, .data$ucn[1]), .data$split_values_2),
      unit = ifelse(.data$to_convert, .data$to, .data$unit),
      value = ifelse(!is.na(.data$split_values_1) & !is.na(.data$split_values_2), paste0(.data$split_values_1, "--", .data$split_values_2), .data$value),
      value = ifelse(!is.na(.data$split_values_1) & is.na(.data$split_values_2), paste0(.data$split_values_1, "--"), .data$value)
    ) %>%
    dplyr::ungroup()  %>%
    dplyr::select(dplyr::any_of(vars))
```
I've looked in detail at building "vanderMoezel_1987", because it requires complex combinations of conversions. The trait "fire_time_from_fire_to_flowering" is input in years (a) and needs to be converted to months (mo). There are a mix of `value_type = bin` and `value_type = mean`. This dataset also has the trait `reproductive_maturity`, also with a mix of `value_type = bin` and `value_type = mean`, but in this case the input and output units are years (a). 

[...]

Issue #10, opened 2023-06-29
Unit conversions for bins


ehwenk:
There is no way to convert units for ranges of values or for bins. For ranges, they are split into minimum/maximum values, but for bins the data must be input in the correct units. Fixing (changing) this would require a function that splits either bins or ranges, performs the conversion, then re-merges the two values. 

There is a recurrent error about a-mo (year to month) conversions when building AusTraits which is most likely due to this. (Although we're been unable to conclusively isolate which study this is due to.

ehwenk:
closed with [1636c31](https://github.com/traitecoevo/traits.build/commit/1636c3154d6ce1853ffd0b601f1eeab516826380)

PR #69, opened 2023-09-15
Changes to units processing

ehwenk:
* add functionality to read in units from a column (issue #7)
* make it possible to have unit conversions for values that are bins/ranges  (issue #10)
* fix warnings during unit conversions (issue #65)
* split function `process_flag_unsupported_values`, so that `unsupported characters` and `does not convert to numeric` errors are assigned before unit conversions and `value out of range` errors are assigned after unit conversions


yangsophieee:
I still need to add tests for this and fix a few tests that are currently failing.

ehwenk:
> I still need to add tests for this and fix a few tests that are currently failing.

I think there is a bigger issue with these changes that I hadn't appreciated that is one of the reason tests are failing - now that units are read in during `dataset_process` rather than `dataset_config` the entire remake file isn't really accurate and although AusTraits builds fine with the empty units_conversions_functions argument, it seems the tests are complaining.

ehwenk:
@dfalster As part of allowing units to be read in from a column instead of just from metadata[["traits"]], the unit conversions file is now required for `dataset_process` instead of `dataset_config`. I've currently shortcut the way the units conversions file was being sources through the remake file, but this should be changed.
- need to remove unit conversions from `dataset_config` - both in the function and in how the remake file is constructed
- need to add unit conversions from `dataset_process` - both in the function and in how the remake file is constructed
- delete line 86 in process.R (currently commented out)
- change line 115 in process.R to reflect change; that is where unit conversions are currently being sourced.

I think I know how to change the building of the remake file, but also a little worried about trying it - and thinking maybe this should wait until austraits.build swaps to traits.build package.

dfalster (approved):
Nice work!

I've simplified the unit conversions for bins and ranges. Can you please check it works on austraits.build?

Issue #54, opened 2023-09-03
Add test ensuring a dataset can pivot wider


ehwenk:
Add a test to `dataset_test()` to check if the following code returns 0 rows. If it doesn't there is a source of duplication somewhere.

_Note, this hasn't been changed to consider the newly revamped `method_id` and `method_context_id`. Once that is merged in, the variable `method_id` below needs to become `method_context_id`, and `method_id` needs to be added to the list, and `15:ncol(.)` changed to `16:ncol(.)`_
```
austraits$traits %>%
  select(dataset_id, trait_name, value, observation_id, source_id, taxon_name, entity_type, life_stage, basis_of_record, value_type,
         population_id, individual_id, temporal_id, method_id, entity_context_id, original_name) %>%
  pivot_wider(names_from = trait_name, values_from = value, values_fn = length) %>%
  pivot_longer(cols = 15:ncol(.)) %>%
  rename(trait_name = name, number_of_duplicates = value) %>%
  select(dataset_id, taxon_name, trait_name, number_of_duplicates, observation_id, entity_type, value_type, population_id, everything()) %>%
  filter(number_of_duplicates > 1) %>%
 nrow()
```

yangsophieee:
Is this a duplicate of issue https://github.com/traitecoevo/traits.build/issues/8?

ehwenk:
I see now that the content of issue #8 is very similar, but I'd assumed from the title it referred to something more specific. I'll close issue #8 - since I know this issue contains the current code that works. And documents the changes needed once the method_id branch is merged in.

yangsophieee:
I've added this to `dataset_test` on branch `test_for_pivoting_wider` but should it also be added to the GitHub actions?

ehwenk:
Eventually yes, but I think it is probably good to test first at the dataset level - to ensure it is only triggered when we expect it to be.

PR #74, opened 2023-09-18
Test for pivoting wider

yangsophieee:
Add Lizzy's test for pivoting wider to `dataset_test`:
- Solves https://github.com/traitecoevo/traits.build/issues/54

I'm wondering if this should also be a test in the GitHub actions as well eventually?


R/testdata.R
```diff
@@ -695,12 +695,42 @@ dataset_test_worker <-
...
+
+        expect_equal(
+          dataset$traits %>%
+          select(
+            dplyr::all_of(c("dataset_id", "trait_name", "value", "observation_id", "source_id", "taxon_name",
+            "entity_type", "life_stage", "basis_of_record", "value_type", "population_id", "individual_id",
+            "temporal_id", "method_id", "method_context_id", "entity_context_id", "original_name"))
+          ) %>%
+          pivot_wider(names_from = "trait_name", values_from = "value", values_fn = length) %>%
+          pivot_longer(cols = 16:ncol(.)) %>%
```
dfalster, line 723: Where does the 16 come from? If from the vector of names above, I suggest adding that as a vector above, then drawing the number from the length of the vector. 

Hard coded numbers are dangerous, as they don't adjust when we change something

dfalster (approved):

R/testdata.R
ehwenk: After pivoting wider column 16 marks the beginning of the columns with traits, while columns 1-15 are the variables above (minus trait_name which becomes the column headers and value). The number of columns with traits is variable dataset by dataset, but as long as you pivot across those variables, they always start at column 16.

PR #80, opened 2023-09-19
Fix error messages

### LINKED ISSUE #5: Change error warning for missing trait name
Referencing [this issue from austraits.build](https://github.com/traitecoevo/austraits.build/issues/728):

In `traits$excluded_data`, if a `trait_name` isn't found, the error is `Missing unit conversion`. It should be changed to be `Trait name incorrect` (or similar).

There are also correct instances of the error `Missing unit conversion`, when it is a missing unit conversion, but the same error is being applied to two different situations.

yangsophieee:
- Fixes #5 
- Added some dummy traits to check whether errors are fixed
- Removed first instance of `not_allowed` as I implemented some newer code but forgot to remove it before merging


dfalster (approved):
This is a helpful change!

PR #82, opened 2023-09-20
Update `furrr` pipeline

yangsophieee:
- Added `workers` argument to `build_setup_pipeline` to allow customisation of `workers` argument during parallel processing
- Added testing of `furrr` method
- Fixed `unit_conversion_functions` argument in `dataset_build`


dfalster (approved):
Great work!

Issue #68, opened 2023-09-13
Add a proper "repeat measurements id"


ehwenk:
There are now two groups who will be using traits.build who require a `repeat measurements` counter to keep track of individual measurements on curve data. The current method of creating replicate counters in custom_R_code, then mapping this in as a temporal context _works_ but is not a good solution if it has to be added to every dataset - not least because it isn't a context and errors can easily be introduced if users don't group by the correct collection of columns.

Instead, it should be automated. Thoughts:
- `repeat_measurements_id` (or some other name) will be added to austraits$traits
- repeat_measurements_id = FALSE will be added to `metadata_create_template` - because otherwise I think this will mask when there should be actual contexts added; in other words, default is to not have repeat_measurement_id's enabled
- if repeat_measurements_id = TRUE, we'll add something to process.R that groups by all the current "need to pivot" columns, then adds a counter if nrow > 1 for any grouping.  Columns are:
```
dataset_id, trait_name, observation_id, source_id, taxon_name, entity_type, life_stage, basis_of_record, value_type,
         population_id, individual_id, temporal_id, method_id, entity_context_id, original_name
```

ehwenk:
A few things here, some duplicated from above notes:

- I started a branch, but I think it is best if NO default changes occur to `metadata_create_template`; can delete those. It just won't appear in the metadata file unless needed.
- People could add `repeat_measurements_id: TRUE` at either the dataset or trait level - at the trait level if the entire dataset is response curves of some sort.
- As a default, within process.R, `repeat_measurements_id = NA`, with row number counters added if flipped to true.
- Need to add repeat_measurements_id to schema (I think true on my branch)
- Requires that in the pivot test, the ncol goes up 1 again, and `repeat_measurements_id` added to list of vars to select and pivot on

yangsophieee:
Resolved with [this commit](https://github.com/traitecoevo/traits.build/commit/9e4ceb94a1c726d3df029077bb2ddc6a9ff65251)

PR #84, opened 2023-10-06
Add `repeat_measurements_id`

yangsophieee:
- Addresses Issue #68 
- Add optional `repeat_measurements_id` to `metadata_create_template`


dfalster (approved):
Nice work!

The only bit that left me somewhat confused were the large number of changes to test datasets. Seems like it's more than is needed for this PR. can you add a note to the PR explaining why so many tests are being added. 

Otherwise you're good to merge

R/process.R
```diff
@@ -151,6 +127,7 @@ dataset_process <- function(filename_data_raw,
   }
 
   # Where missing, fill variables in traits table with values from locations
+  # Currently overwriting dataset-level column metadata -- NEED FIX
```
ehwenk, line 130: Isn't that what is meant to happen? If there is location-specific metadata, it should overwrite the dataset-level value for observations at that location.

yangsophieee: @ehwenk It's overwriting the dataset-level metadata which is read in from a column. I thought the idea was that if it's read in from a column, it's likely the final value for that observation.

ehwenk: @yangsophieee Oh I see, I thought you just meant it was overriding a fixed value. I see the problem though - because you have a add code that has different actions for fixed values vs values from columns. Maybe only allow the location-level or trait-level value to be used if the dataset value is `fixed` (i.e. not a column name)?

yangsophieee:
Resolves #68

Issue #87, opened 2023-10-06
`dataset_test` picks up time values as unallowed


yangsophieee:
Currently `dataset_test` fails when substitutions for time variables (e.g. `flowering_time`, `fruiting_time`) exist, because of [this line](https://github.com/traitecoevo/traits.build/blob/53d56ef18051e38808ecf6ee7e3e261e3f38307e/R/testdata.R#L666) looking for whether values are found within the trait definitions. 

I can think of several easy ways to fix this: if the values only contain "n" and "y" then you don't check for the trait definitions. 

This made me think about [this line](https://github.com/traitecoevo/traits.build/blob/53d56ef18051e38808ecf6ee7e3e261e3f38307e/R/process.R#L1024) in `process.R` which specifically codes for the traits `flowering_time` and `fruiting_time`. But this isn't very generalisable to other databases if they want to make their own time traits. I wonder if it's worth making a new "time" category of traits, in addition to "categorical" and "numeric". What do you think @ehwenk?

ehwenk:
I've previously thought about having a separate trait category for `flowering_time`, etc. for similar reasons. Because when I've added a new time-trait (`recruitment_time`), you're right, I then had to tweak code (I think for building the dictionary ontology). And it probably should have been added to this list as well. 

The specific test you referenced searches for `N`'s and `Y`'s only, which is specific to time. 

But a separate problem with the `time` traits is that they are categorical traits for which the allowable trait values aren't listed. That could be true to non-time traits - given the trait `recruitment_time` isn't causing any errors, it seems we've already adapted the traits.build workflow to handle instances where categorical traits don't have set values. Maybe this is actually only a problem for the code to build the APD, in which case I guess it is fine to require a fix.

@dfalster what do you think? Is it a `time` category we should consider adding or a `factor` (or other word) category for traits without set values. Or nothing?

dfalster:
Backlog triage: **the stopgap landed, the actual design question is still open, and there is a concrete piece of AusTraits-specific code left in the generic package.**

**What was done:** the narrow fix the issue itself suggested. `R/testdata.R:653` now excludes 12-character Y/N strings from categorical-definition checking:

```r
to_check <- to_check[!(grepl("^[YyNn]+$", to_check) & stringr::str_length(to_check) == 12)]
```

[...]

Issue #49, opened 2023-08-30
Time data types get reformatted with `read_csv`


yangsophieee:
Time data types get reformatted with `read_csv`, such that e.g. "9:00:00" becomes "09:00:00". This is not an issue if it's consistent with the output of `metadata_add_contexts` and `process.R`. However, right now `metadata_add_contexts` recognises "09:00:00" but then it gets written to the metadata as "32400.0". 

![image](https://github.com/traitecoevo/traits.build/assets/78716301/a90063b1-0278-4583-abf6-3c631caca0fc)
![image](https://github.com/traitecoevo/traits.build/assets/78716301/1151be84-fb06-46d8-9950-b645de64a40e)

yangsophieee:
Note: You can't read in all data as character columns during `read_csv` because then `process_custom_code` won't work. 

I decided to convert context values to character type before they are converted to yaml, which seems to have fixed the problem!

yangsophieee:
There is a little bit of an issue where someone might manually add a time context value to the metadata in its original format, i.e. "9:00:00". Then they would find that the context value isn't being detected in the data (because the data has been read in as "09:00:00". There is a `dataset_test` that says when context values are not in the metadata so this would clear things up, but it's still a bit confusing.

ehwenk:
Nice solution. Does the dataset_test now specify which values `data` are missing from `metadata` vs which values `metadata` values are extra? If so, I think that is good enough, because it will immediately indicate what to change.

yangsophieee:
@ehwenk Not yet, that is still an issue to be fixed (#63) but I will get to it soon!

yangsophieee:
It seems that David has been converting time variables to character type within `custom_R_code`, otherwise the `dataset_test` test about time data types gets flagged. Needs further investigation, but it seems "11:30" in the data file may not be properly written by `metadata_add_context` to the metadata file.

dfalster:
Fix in #220, which covers both halves of this.

**The part that was missed:** 96b6abc added the `as.character()` coercion to the *interactive* branch of `metadata_add_contexts()` only. The identical line in the `user_responses` branch was left untouched, so that path still writes seconds. Reproduced on a copy of `Test_2022` with a time column:

```yaml
values:
- find: 32400.0
  value: unknown
```

[...]

PR #88, opened 2023-10-06
Fix time data types in contexts getting reformatted

yangsophieee:
Addresses #49
- Time data types are getting reformatted with `read_csv`, because it detects the column data type when reading in, such that e.g. "9:00:00" becomes "09:00:00". This is not too much of an issue if it's consistent with the output of `metadata_add_contexts` and `process.R`. However, `metadata_add_contexts` was rewriting "09:00:00" as total number of seconds: "32400.0"
- I've fixed it so that context values are [character type](https://github.com/traitecoevo/traits.build/blob/53d56ef18051e38808ecf6ee7e3e261e3f38307e/R/setup.R#L482) so that [as.yaml](https://github.com/traitecoevo/traits.build/blob/53d56ef18051e38808ecf6ee7e3e261e3f38307e/R/utils.R#L263) in `write_metadata` does not convert times to number of seconds
- See [this comment](https://github.com/traitecoevo/traits.build/issues/49#issuecomment-1750121051) about the rewriting from "9:00:00" to "09:00:00" (I've added a note to the user when time data types get reformatted)
- Also added a fix for some of issue #87, where substitutions for time variables (e.g. flowering_time, fruiting_time) fail `dataset_test`


yangsophieee:
@dfalster This is ready for review :) Thank you!

dfalster (approved):

Issue #55, opened 2023-09-06
Add test to `dataset_test` ensuring `collection_date` follows `custom_R_code`


yangsophieee:
Add test to `dataset_test` ensuring `collection_date` follows `custom_R_code`, because `read_metadata` relies on `collection_date` to find the end of `custom_R_code`: https://github.com/traitecoevo/traits.build/blob/383f6fa56d219f41b96f7014206aa06d5d952526/R/utils.R#L216

Is there a better way to find the end of `custom_R_code`, like looking for `'` on its own line or at the end of a line? The issue is if there's ever a case where `'` ends a line but more code follows.

yangsophieee:
Closed with [this commit](https://github.com/traitecoevo/traits.build/commit/b76fd3b0f3012c80d34b8ebd4449b279407c85d5)

PR #89, opened 2023-10-06
Add test for `collection_date` following `custom_R_code`

yangsophieee:
Addresses issue #55 
- Also wondering if there's a more elegant solution for preserving `custom_R_code` formatting that doesn't need to look for `collection_date`


dfalster (approved):
Good idea!

PR #92, opened 2023-10-11
Solving minor issues

yangsophieee:
- Add Sophie to DESCRIPTION file
- Add a package prefix to avoid R CMD Check warning
- Remove unnecessary line from `metadata_create_template` as requested by @ehwenk


ehwenk (approved):
Thanks for these corrections

Issue #94, opened 2023-10-12
NAs in an `individual_id` column lead to NAs in final `individual_id` column


yangsophieee:
Currently if you specify an `individual_id` column in the metadata, `individual_id` will be NA in the final traits table if there are NAs in the specified metadata column. I found these comments in process.R which suggests the goal was that `individual_id` would still be generated for the NAs:

```
  # For a file where `individual_id` is specified in the metadata,
  # if there are rows of data where the `individual_id` is `NA`,
  # these are filled in with the row_number(),
  # just as would occur if no `individual_id` column is specified
```

Looking into fixing this now.

Issue #90, opened 2023-10-09
Add `collection_date` to variables that make `observation_id`? 


yangsophieee:
For the invert database there are a few cases where `collection_date` is used for repeat measurements. I was wondering whether we should add a "replicate observations" context in these situations, or whether we should end up trying to add `collection_date` to the pivot variables. Previously Lizzy had concerns that it might not work as a pivot variable because it is continuous and sometimes includes time. I'm imagining it would be okay if it were all character type?

@ehwenk's thoughts about adding it:

> I'm trying to reason through the best answer for `collection_date.` In particular, how the "meaning" of `collection_date` changes if it is an "observation_id-distinguisher". Then it would need to be added to the list of variables that generates `observation_id` - because it isn't just about pivoting, but more about how it informs what is an observation. And I think that is fine - in fact better. In my ontology thought-charts, I do have `collection_date` as an alternative to `temporal_context` for informing about "time", but the way it was input, it didn't actually get used, but just "reported".
> The one complication comes if there are trait-level `collection_date` entries. So far it is always dataset-level, although I know we talked about allowing this. If there are different `collection_date` columns for different traits, then they'd become separate `observation_id`'s, but retain the same `individual_id` (since that would be based on the row), so I think that is fine.

yangsophieee:
More thoughts from Lizzy:

> I just thought of a problem with including collection_date in observation_id!! For response curve data, there is a time column that is of course different for each point - and then those will become different observations. 

> I guess since this is an expert use case, I'll simply add a note for these to ensure the column mapped to collection_date is not the same as the time stamp for each measurement.

Time stamp of repeat measurements is usually ignored (not added as a context) because otherwise it would make each repeat measurement a different `observation_id`.

yangsophieee:
Also include `original_name`

PR #95, opened 2023-10-12
Add `collection_date` and `original_name` to `observation_id`

### LINKED ISSUE #67: Update schema for reading in metadata from different sections
Right now `dataset_test` doesn't allow `basis_of_value`, `value_type` to be read in from the dataset level and `collection_date` from the trait level (e.g. in Test Dataset 1). I'm sure there's other metadata fields to review, which we may or may not want to be read from these different levels.

yangsophieee:
- Fixes #67, #90 and #94
- Also fixes an issue I found with `unit_in`/`unit` where the column name instead of the column value was being read in whenever the `unit_in` value was NA (discussed with Lizzy)
- Added a hard fix to make sure categorical traits have NA units


dfalster (approved):
Complex stuff, well done!

ehwenk (approved):
Well done working out that the solution to the units problem was to remove lines of extraneous code!

Issue #96, opened 2023-10-15
`exclude_observations` field not working


ehwenk:
The metadata[["exclude_observations"]] field is currently not causing data to be moved to the `excluded_data` table. This is something that has been broken recently (in the past 1-2 weeks). 

I initially was looking at the austraits.build study `Fonseca_2000` and wondered if it was because the names listed are actually `originial_names` (they are all capitalised as part of automatic processing), but then built all of AusTraits and can confirm there are no values appearing in austraits$excluded_data with the error "Observation excluded in metadata" .

```
> austraits$excluded_data %>% filter(error == "Observation excluded in metadata")
# A tibble: 0 × 27
```

ehwenk:
It looks like the change occurred with commit https://github.com/traitecoevo/traits.build/commit/119502867a6ce5b3d914cd9c78101f7e9843af33

Previous code, Line 659 Process.R:
```
dplyr::mutate(error = ifelse(data[[v]] %in% fix[[v]]$find, "Observation excluded in metadata", .data$error))
```

Current code, Line 881 Process.R:
```
      dplyr::mutate(
        error = ifelse(.data$trait_name == v & .data$value %in% fix[[v]]$find,
        "Observation excluded in metadata", .data$error))
```

dfalster:
How could I easily trigger this?

dfalster:
never mind, I see, `Fonseca_2000`

yangsophieee:
Closed with this commit: https://github.com/traitecoevo/traits.build/commit/dda867348f2a9d391f4ae52d4515f0c86be3567f

PR #97, opened 2023-10-16
Fix error with exclusions (#96)

dfalster:
This fixes issue with data not being excluded, as described in #96

@ehwenk noted a change with this commit https://github.com/traitecoevo/traits.build/commit/119502867a6ce5b3d914cd9c78101f7e9843af33
but the original code was correct. So have reverted back to the original.

some tests are now failing, as the dataset length is different to what it was when the test dataset was generated. 

@yangsophieee - we need to generate data for `Test_2023_1`. Can you do this?

Can you also add instructions into the README for each test dataset?


ehwenk (approved):
I've built AusTraits from this branch and have confirmed that the expected exclusions are in `austraits$excluded_data` again

yangsophieee (approved):
I've regenerated expected outputs, updated README and fixed a typo in metadata for `Test_2023_1`

Issue #60, opened 2023-09-06
Precedence of trait-level metadata over location-level metadata not working


yangsophieee:
Currently location-level metadata is overwriting trait-level metadata, e.g. for Test_2023_2 - Alstonia scholaris, `leaf_mass_per_area` should have `basis_of_value` as `measurement` but it's currently `expert_score` (the value for the location Cape Tribulation). 

I think this is occurring in the loop around [Line 161](https://github.com/traitecoevo/traits.build/blob/3ba84a9937956d64fd186c69526fff5f40583fc5/R/process.R#L161) of `process.R`. It's happening for both long and wide datasets

ehwenk:
This is actually intentional - we had to code either trait-level or location-level metadata to take precedence over the other and fairly randomly picked location. If you have a good reason to swap it around to trait-level metadata taking precedence, you're welcome to change it.

ehwenk:
@yangsophieee I am very happy for you to switch it to trait metadata > location metadata > dataset metadata!

yangsophieee:
@ehwenk thanks! what about the column precedences?

ehwenk:
@yangsophieee I think it probably would always be the same - you shouldn't fill in anything at the trait_level if there is a universal value or column given at the dataset level. Filling in at the trait_level (or location_level) should be a clear signal that this is the correct variable value/column for that trait/location

ehwenk:
@yangsophieee I hadn't thought about the different workflows for fixed values vs values from columns. 

For fixed values, we decided values specified in the traits-section overwrites values from the dataset-section; and location-section values overwrite traits-section values. (Although I have no preference actually for the order or traits vs location.)

Meanwhile dataset-level column-specified values should have ultimate precedence - otherwise what does it mean to read in a value from a column if you're going to overwrite it? Then that cell in the data.csv file should be edited (either actually edited, or using custom_R_code). Is it realisitic to add an `ifelse` that only allows the location-level or trait-level overwriting of values if the dataset-level is a fixed value?

ehwenk:
@yangsophieee

I've just spent a long time staring at Test_2023_4 and Test_2023_2 outputs and am not sure anything needs to be changed. As in, I don't think issue #60 and issue #38 are issues. 

What are specific variables, datasets that you think are giving erroneous outcomes?

I also think that I misstated something in my last comment.

In order of information being read in (i.e. lower in this list overwrites items at top):

1. Read in dataset-level fixed value
[...]

PR #99, opened 2023-10-18
Check precedences of metadata

yangsophieee:
Addresses #60, checking precedence of metadata fields across various sections of metadata are working correctly (e.g. location > trait metadata)
- Added additional testing for checking precedence of metadata
- Fixed an error with `dataset_test` where I didn't realise the `data` variable already uses `process_custom_code` so it was applying it twice
- Removed old testing files
- Small formatting changes
- @ehwenk and I agree that trait metadata should probably take precedence over location metadata but we will leave this to a later date (trait metadata is read in via `process_parse_data` and location metadata replaces it afterwards in `dataset_process` around Line 141, so you'd have to move the location metadata part into `process_parse_data`, which will presumably require splitting location data into location properties vs the other variables and having them input at different times)


dfalster:
Let me know once passing and I can review!

yangsophieee:
@dfalster it's passing now! Thanks :)

dfalster (approved):

yangsophieee:
Addresses #60, checking precedence of metadata fields across various sections of metadata are working correctly (e.g. location > trait metadata)
- Added additional testing for checking precedence of metadata
- Fixed an error with `dataset_test` where I didn't realise the data variable already uses `process_custom_code` so it was applying it twice
- Removed old testing files
- @ehwenk and I agree that trait metadata should probably take precedence over location metadata but we will leave this to a later date (trait metadata is read in via `process_parse_data` and location metadata replaces it afterwards in `dataset_process` around Line 141, so you'd have to move the location metadata part into `process_parse_data`, which will presumably require splitting location data into location properties vs the other variables and having them input at different times)

PR #100, opened 2023-10-19
Fix unit conversions warning

yangsophieee:
The unit conversions warning (#65) was showing up again for both @ehwenk and me. It seems to me that there was something going on with the `mutate` `ifelse` statement not being vectorised in the expected way. I don't fully understand why it wasn't working as expected (because it should've been) but after several hours of testing I've just decided to accept that adding `rowwise` works and produces the expected output.

The problem dataset was once again, vanderMoezel_1987.

### DIFF
--- a/R/process.R
+++ b/R/process.R
@@ -1196,6 +1196,7 @@ process_convert_units <- function(data, definitions, unit_conversion_functions) 
   # Split by unique unit conversions, to allow for as few calls as possible
   data <- data %>%
     dplyr::group_by(.data$ucn, .data$to_convert) %>%
+    dplyr::rowwise() %>%
     dplyr::mutate(
       # Standard conversion
       value = ifelse(.data$to_convert == TRUE &                     # Value requires conversion
@@ -1203,8 +1204,8 @@ process_convert_units <- function(data, definitions, unit_conversion_functions) 
                       !is.na(.data$value),                          # Value not NA - the full matrix from data.csv file is still in data table
                       f_standard(.data$value, .data$ucn[1]),        # Convert value to appropriate units
                      .data$value),                                  # If conditions not met, keep original value
-      # Value is a range of bin
-      value = ifelse(.data$to_convert == TRUE  &                    # Value requires conversion
+      # Value is a range or bin
+      value = ifelse(.data$to_convert == TRUE &                    # Value requires conversion
                       .data$value_type %in% c("bin", "range") &     # `value_type` is a bin or range
                       !is.na(.data$value),                          # Value not NA - the full matrix from data.csv file is still in data table
                       f_range_bin(.data$value, .data$ucn[1]),       # Convert value to appropriate units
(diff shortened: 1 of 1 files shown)


dfalster:
I'll take a look at this before merging

R/process.R
```diff
@@ -1188,28 +1188,30 @@ process_convert_units <- function(data, definitions, unit_conversion_functions)
...
-    purrr::map(f_standard, name) %>%                       # apply unit conversions to each
-    purrr::map_chr(~paste(.x, collapse = "--"))   # paste back together
+    stringr::str_split(value, "\\-\\-") %>%       # Split into parts
+    purrr::map(f_standard, name) %>%              # Apply unit conversions to each
+    purrr::map_chr(~paste(.x, collapse = "--"))   # Paste back together
   }
 
   # Split by unique unit conversions, to allow for as few calls as possible
   data <- data %>%
     dplyr::group_by(.data$ucn, .data$to_convert) %>%
```
dfalster, line 1198: need to add value_type to group_by call,  ie

```
dplyr::group_by(.data$ucn, .data$to_convert, .data$value_type)
```

R/process.R
```diff
@@ -1188,28 +1188,30 @@ process_convert_units <- function(data, definitions, unit_conversion_functions)
...
+    dplyr::rowwise() %>%
```
dfalster, line 1199: move rowwise as will make everything much slower

R/process.R
dfalster, line 1198: need to add value_type to group_by call,  ie

```
dplyr::group_by(.data$ucn, .data$to_convert, .data$value_type)
```

dfalster (changes requested):
I looked into this. The issue can be resolved by adding value_type into the group_by statement. 

When I checked what was happening inside the functions `f_standard` binned data was still being fed into this. Adding value_type into the group_by fixes this. 

This is preferable to rowwise, as rowdies will result in more function calls

yangsophieee:
@dfalster Thanks Dan!!! I was wondering what debugging tools you used to look into what was being fed into the functions? I struggle with loops where only one particular iteration throws a warning.

dfalster:
In this case I put a browser() call first in `process_unti_conversions`, then also in `f_standard` -- with the latter I was able to check the values being called within `f_standard` , which is when I saw values that were bins and shouldn't be there. 

This would cause lots of browser calls, so I only ran this specifically on the problem study: vanderMoezel_1987

yangsophieee:
Oh ok! I did do that and saw the bins getting read in but I thought the ifelse function would only run f_standard on the ones that aren't bins.

dfalster:
me too. hence why I double checked with a browser within `f_standard`

yangsophieee:
@dfalster Changes have been implemented! Thank you :)

dfalster (approved):

PR #102, opened 2023-10-19
Generalised database and target names

### LINKED ISSUE #6: Read in `entity_type`, `value_type`, etc. from columns for long data formats
Referencing [issue from austraits.build](https://github.com/traitecoevo/austraits.build/issues/725):

@ehwenk pointed out that when in long format, we might want to read in columns of data for things like `entity_type` and `value_type,` and that currently this isn't possible. The issue was fixed for wide format.

yangsophieee:
Closes #6
- Allows different databases to rename the database, instead of before using "austraits"
- Default is just set to "database", @ehwenk and I decided against "traits" because "traits$traits" is confusing


yangsophieee:
Need to fix test-setup.R

dfalster (approved):
Super!

PR #103, opened 2023-10-19
Create test for blank datasets

### LINKED ISSUE #4: Create more explicit error message for "blank" datasets
Referencing [this issue from austraits.build](https://github.com/traitecoevo/austraits.build/issues/732) opened by @ehwenk:

When there is a dataset that contains no information, usually due to a mistake in trait names being mapped in, you get this error:

```
Error in `dplyr::bind_rows()` at austraits.build/R/process.R:1391:4:
! Can't combine `..1$taxon_name` <character> and `..332$taxon_name` <logical>.
```
The empty dataset can be identified by `dir("data")[332]`

It would be good if you got an error message that had this error pattern, the error explicitly identified the problem for traits.build users outside of the AusTraits team.

yangsophieee:
Closes #4 
- These changes address two scenarios: 1) all traits in the metadata have `trait_name` mapped to `.na`, leading to an empty traits table, 2) all traits have 'wrong' `trait_name`'s or `trait_name`'s not in the traits dictionary, leading to an empty `traits` table
- Picks up these issues with `dataset_test`

@ehwenk Maybe you can do some testing with empty datasets to check that it works as expected?

### DIFF
--- a/R/process.R
+++ b/R/process.R
@@ -680,7 +680,7 @@ process_create_context_ids <- function(data, contexts) {
 
   # Extract context columns
   context_cols <- data %>%
-    dplyr::select(dplyr::all_of(tmp$var_in)) %>%
+    dplyr::select(dplyr::any_of(tmp$var_in)) %>%
     dplyr::mutate(dplyr::across(everything(), as.character))
   names(context_cols) <- tmp$context_property
 
--- a/R/testdata.R
+++ b/R/testdata.R
@@ -745,18 +745,6 @@ dataset_test_worker <-
                       ))
         }
 
-        ## Check that dataset can pivot wider
-        expect_no_error(dataset <- test_build_dataset(
-          file.path(path_data, dataset_id, "metadata.yml"),
-          file.path(path_data, dataset_id, "data.csv"),
-          dataset_id,
-          get_schema("config/traits.yml", "traits"),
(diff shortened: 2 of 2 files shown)


dfalster (approved):

ehwenk:
And there is a third scenario - the units are so far off that all data are excluded.

I've tested these scenarios and all three give warnings - either:

units way off:
```
Failure: Togashi_2015
nrow(dataset$traits) == 0 is not FALSE

`actual`:   TRUE 
`expected`: FALSE
Togashi_2015 - `traits` table is empty
```

all NA's for trait names:
```
Failure: Jurado_1992
... == 0 is not FALSE

`actual`:   TRUE 
`expected`: FALSE
data/Jurado_1992/metadata.yml - `traits` metadata only contains NA `trait_name`'s

[ FAIL 1 | WARN 0 | SKIP 0 | PASS 98 ]
```

all trait names not in dictionary:
```
Failure: Jurado_1992
data/Jurado_1992/metadata.yml - traits - should not contain: seeds

Failure: Jurado_1992
nrow(dataset$traits) == 0 is not FALSE

`actual`:   TRUE 
`expected`: FALSE
Jurado_1992 - `traits` table is empty
```

ehwenk (approved):
As described in another comment, I've tested the scenarios likely to lead to blank traits tables and they are all caught by this test - so perfect.

Issue #101, opened 2023-10-19
Reports not building


ehwenk:
Something small is not working with the reports building - in the categorical traits section.

To be explored further.

dfalster:
I thought there was test for this function?

ehwenk:
I talked to Sophie and there is - once I work out what is wrong, I'll figure out why the test didn't pick it up

ehwenk:
Closed with [4a1674d](https://github.com/traitecoevo/traits.build/commit/4a1674d45dbd9f6d2b0eed1487c3367f7525f83e)

PR #108, opened 2023-10-26
Update report_dataset.Rmd

ehwenk:
Fix error in renaming syntax introduced to reports at some point since June.

Solves issue #101


yangsophieee (approved):

yangsophieee:
Thanks for picking up the mistake! The `all_of` and tidyselect naming conventions are tough.

PR #109, opened 2023-10-27
Improve `dataset_test` failure messages

### LINKED ISSUE #48: How do you suppress default failure messages from `expect_` functions?
To make `dataset_test` more accessible, I want to suppress default failure messages from `expect_` functions and only show the message that we have manually designed to the user.

image

E.g. We don't want the code and the `actual` and `expected` lines  to show. I tried looking into the customisation of error messages within the `expect_` documentation but couldn't figure it out. 
For example, in `expect_equal`:
- `object` and `expected`: "Both arguments supports limited unquoting to make it easier to generate readable failures within a function or for loop. See [quasi_label](vscode-webview://0himrfqpbn9qvtmpkamgvvl4gnieo9dm953pmjqq9uiegpb1tpu4/testthat/help/quasi_label) for more details."
-  `label`: "Used to customise failure messages. For expert use only." 

I tried customising these but it doesn't come out the way I want. In addition, the `info` argument is what we've been using and it's been soft-deprecated. I can't figure out why.

yangsophieee:
Closes #48 
- Removed the duplicate set of failure messages by switching from `default_reporter()` (ProgressReporter) to CompactProgressReporter
- Swapped out `testthat` `expect_` functions for manually written expectations in order to fully customise failure messages and make them more accessible
- Tidied all failure messages
- Flagged an issue in the `testthat` GitHub page to add an option to turn off backtraces for wrapped expectations (again to improve accessibility)


R/testdata.R
```diff
@@ -55,91 +55,123 @@ dataset_test_worker <-
 
     # We're using 2nd edition of test that, which has "context" field
     # https://cran.r-project.org/web/packages/testthat/vignettes/third-edition.html
+    # Is there a reason to be using the 2nd edition, while context has been superseded in later versions?
```
dfalster, line 58: I fixed the edition as updating would require several tests to be updated and I had limited time available. You can update if you want. But it's also not necessary.

R/testdata.R
```diff
@@ -55,91 +55,123 @@ dataset_test_worker <-
...
     }
 
-    expect_not_NA <- function(object, info = NULL, label = NULL) {
+    expect_equal <- function(object, expected, info) {
+      i <- object == expected
+      comp <- compare(all(i), TRUE)
+      expect(comp$equal, info)
+    }
+
+    expect_true <- function(object, info) {
```
dfalster, line 114: Aren't these already included in testthat?

R/testdata.R
```diff
@@ -208,87 +240,77 @@ dataset_test_worker <-
...
-        expect(is.null(error),
-               sprintf(
-                 "%s threw an error: %s",
-                 label,
-                 paste(error$message, collapse = ",")
-               ),
-               info = info)
-        invisible(NULL)
-      }
+    expect_no_error <- function(object, regexp = NULL, ..., info) {
```
dfalster, line 246: Looks like regex isn't being used?

dfalster (approved):
Looks good @yangsophieee , but I haven't run it. Please ask @ehwenk to try it out before merging. If she's ahoy then feel free to merge

R/testdata.R
yangsophieee: Ah ok that's fine then! It works fine.

R/testdata.R
yangsophieee: Yes, but writing my own versions allow full customisation of failure messages, otherwise there's a lot of fluff in the defaults.

R/testdata.R
yangsophieee: Yeah it's not being used, I didn't add it so happy to get rid of it.

PR #111, opened 2023-10-27
Taxonomy documention & edits

ehwenk:
A very trimmed down version of `build_update_taxonomy` that requires a taxon_list.csv file with just 3 columns, `taxon_name`, `cleaned_name1`, and `taxon_rank`.

All manipulations to names that previously occurred as part of `build_update_taxonomy` will now be transferred to `update_taxon_list` functions that can be individualised for different studies. 

This branch is ready to review, BUT the AusTraits taxon_list.csv file is no longer in the correct format. AusTraits will build, but all `genus sp.` taxa will currently show up as being unknown taxon names.


dfalster (approved):
Looks promising! 

I'm fine with approach of altering tests to reflect simplified input.

You can leave some extra cols in the schema if you wish

PR #121, opened 2023-11-10
Remove dependency on austraits by importing plot

dfalster:
We're shifting to remove dependency on austraits package. 
We'll import some more functions into this repo  but starting with this as only one required to execute


yangsophieee (approved):
Seems like `forcats` is a dependency of the austraits plot function and needs to be imported.

yangsophieee:
@dfalster Great! I've gone through and fixed a lot of R CMD check notes for "no visible binding"/"no visible global function definition" on the `more_dataset_tests` branch just so you know.

dfalster:
Thanks @yangsophieee !

Oops, I just fixed some of those too in #122. Either, I close that PR or we merge now anyway.

yangsophieee:
@dfalster No it's okay, you can merge it!

Issue #112, opened 2023-10-31
`exclude_observations` does not allow trait values to be excluded


yangsophieee:
Currently `exclude_observations` does not allow trait values to be excluded. It only allows `taxon_name` and other metadata fields to be excluded. You can see this happening in [this code](https://github.com/traitecoevo/traits.build/blob/33dd832b0862f5c646c10e73d006a3640731a0fe/R/process.R#L890), where it looks for a column `v` in `data`. The column doesn't exist if it's a `trait_name`, since at this point `data` has gone through `process_parse_data` and is in long format (hence all trait names are in the `trait_name` column). 

We did not pick this bug up previously because the only excluded observations in AusTraits are for `taxon_name`. I realised the bug when I tried excluding trait value observations in a test dataset. I've fixed the issue on branch [`add_austraits_functions_to_dataset_test`](https://github.com/traitecoevo/traits.build/blob/563aa6548ce1f3afec191ae0676132dad9ee2eb9/R/process.R#L890) and also added testing in `dataset_test`.

yangsophieee:
Fixed with this [PR](https://github.com/traitecoevo/traits.build/pull/123) and [commit](https://github.com/traitecoevo/traits.build/commit/9a9be4c4450406e6fb20966d6d5184bbf5a2f77d).

PR #123, opened 2023-11-13
Add `austraits` functions to `dataset_test` and allow trait values to be excluded in `exclude_observations`

### LINKED ISSUE #93: Make `austraits` package functions part of workflow or testing
Brought to my attention by @ehwenk:

Currently there is no testing to check that the `austraits` functions `join_...`, `as_wide_table`, etc. work on the austraits output. This would be useful, as there have been previous releases where there have been issues in the data preventing `as_wide_table` from working.

@ehwenk Thinking about it a bit more, I'm a bit confused how this would work. Is the aim of this to make sure the `austraits` functions work as intended, or whether the data doesn't have issues in it, assuming that the `austraits` functions work as intended?

If this is added as a test within `traits.build`, it would only test whether the `austraits` functions work on the test datasets. In order to test if a final database output doesn't have any issues in it, couldn't we just run `dataset_test()` on the whole database and check no failures + everything pivots wider?

yangsophieee:
Fixes #93, #112:
- Moved `austraits` functions `trait_pivot_wider` and `trait_pivot_longer` to traits.build and added testing within `dataset_test`
- Moved code for testing pivot wider into standalone function called `check_pivot_wider()`
- Allow trait values to be excluded in `exclude_observations`, where previously only metadata fields/locations, etc. could be excluded
- Added extra tests in `dataset_test` for `exclude_observations` and `taxonomic_updates` sections


NAMESPACE
```diff
@@ -33,6 +34,8 @@ export(metadata_update_taxonomic_change)
 export(plot_trait_distribution_beeswarm)
 export(read_csv_char)
 export(read_metadata)
+export(trait_pivot_longer)
```
ehwenk, line 37: @dfalster What should be the naming convention for these functions? `trait_table_pivot_wider`?

R/pivot.R
```diff
@@ -0,0 +1,139 @@
...
+#' values_long <- trait_pivot_longer(traits_wide)
+#' }
+#' @author Daniel Falster - daniel.falster@unsw.edu.au
+#' @author Fonti Kar - fonti.kar@unsw.edu.au
+#' @export
+trait_pivot_longer <- function(wide_data) {
+
+  wide_data %>%
+    tidyr::pivot_longer(
+      cols = 20:ncol(.),
```
ehwenk, line 134: I think in the documention we should describe where the `20` comes from, because eventually we'll forget that this is the number of "non-trait-value" columns in the wide dataset.

yangsophieee: Good idea, added a comment

R/testdata.R
```diff
@@ -253,19 +253,36 @@ dataset_test_worker <-
...
         expect_allowed(names(object[[i]]), allowed, info = paste(info, i), label = "field names")
-      invisible(NULL)
+      invisible(object)
+    }
+
+    expect_list_elements_exact_names <- function(object, expected, info) {
+      for (i in seq_along(object))
+        expect_contains(names(object[[i]]), expected, info = paste(info, i))
+        expect_allowed(names(object[[i]]), expected, info = paste(info, i), label = "field names")
+      invisible(object)
```
ehwenk, line 285: I don't know why these changes are necessary - can you explain?

R/testdata.R
```diff
@@ -708,16 +725,12 @@ dataset_test_worker <-
         ## TODO do the same for `taxonomic_updates` and `exclude_observations`?
```
ehwenk, line 725: Can you get rid of this line now?

yangsophieee: Done

R/testdata.R
yangsophieee: It's not super necessary but it was to make the code more compact. Instead of doing both `expect_list_elements_contains_names` and `expect_list_elements_allowed_names` on the same object, now you just run one function (`expect_list_elements_exact_names`)

NAMESPACE
dfalster: Comes back to the discussion we were have the other day. What's the prefix for any function using the results of a {traits.build} pipeline. How about `db_traits_pivot_longer`?

R/pivot.R
```diff
@@ -0,0 +1,139 @@
...
+    dplyr::summarise(n_value_type = length(unique(.data$value_type))) %>%
+    arrange(.data$observation_id) %>%
+    dplyr::filter(.data$n_value_type > 1)
+
+  if (nrow(check_value_type) > 1) {
+
+    traits %>%
+      select(-dplyr::all_of(metadata_cols)) %>%
+      # Sophie - what's the point of the group_by?
+      dplyr::group_by(
```
dfalster, line 80: Maybe because the ids aren't unique? so only makes sense to do dataset by dataset?

R/pivot.R
```diff
@@ -0,0 +1,139 @@
...
+#' values_long <- trait_pivot_longer(traits_wide)
+#' }
+#' @author Daniel Falster - daniel.falster@unsw.edu.au
+#' @author Fonti Kar - fonti.kar@unsw.edu.au
+#' @export
+trait_pivot_longer <- function(wide_data) {
+
+  wide_data %>%
+    tidyr::pivot_longer(
+      cols = 20:ncol(.), # This is the start of the trait columns in wide format
```
dfalster, line 134: How can we detect the column to start with, with hard coding the number 20?

dfalster (approved):
Nice work team! some queries to address

R/pivot.R
yangsophieee: The IDs should be unique I think. I tried omitting the `group_by` and didn't see anything different with the output, but would like someone else to try and verify.

R/pivot.R
yangsophieee: I updated the comment to:

```
# This is the start of the trait columns in wide format
# The previous columns are standard metadata and ID fields that do not change
# unless structural changes to traits.build occur
```

Should I instead update the code so that it looks for the column after "original_name", or looks for the first column that isn't in the traits table schema @dfalster?

ehwenk: I've thought about doing that several times, and that will work as long as we remember to always have `original_name` as the final column. So, yes, let's code `1 column past original name` and then put in the notes that is what we've coded, so when something breaks in the future we know the culprit

yangsophieee: @ehwenk In coding that, I realised the number should have been 21, not 20!

R/pivot.R
ehwenk: I think what's going on is that if you're maintaining columns that aren't the pivot, names_from or values_from columns you need to identify the pivot columns with group_by - but because now we're removing all the others there won't be a difference here.

yangsophieee: I tried without the `group_by` and it has the same number of columns with and without -- including columns that aren't needed for the pivot (redundant variables like `entity_type`, `population_id`, etc. The [only columns we remove](https://github.com/traitecoevo/traits.build/blob/24d9927f20128e7135e067b385c78f4e25d19eef/R/pivot.R#L78) are `c("unit", "replicates", "measurement_remarks", "basis_of_value")` right?

ehwenk (approved):
Looks good now

PR #125, opened 2023-11-15
Collapse duplicate categorical values within a row

### LINKED ISSUE #120: Duplicate categorical values within a combination of trait values in a single row
I found an "annual annual ephemeral" value in the `life_history` data of AusTraits v4.2.0. I thought I fixed it a while back so that these would be collapsed to "annual ephemeral" in process.R. Maybe there were issues with automatic substitutions across combinations of trait values. Not sure if this should be in `dataset_test` or `process.R`.

yangsophieee:
Fixes #120
- Also updated `taxon_list.csv` and `taxon_list-orig.csv` - those are all the expected output changes except for Test_2023_1 where "mallee tree shrub tree mallee" gets collapsed to "mallee shrub tree"


ehwenk (approved):
This looks good - but I'm wondering if all the changes to the taxon lists need to also be on this branch. I just worry that there will be conflicts with the next branch where all the "same" changes exist and then you'll have these huge changes to csv files multiple times.

yangsophieee:
@ehwenk You're right! I've reverted the changes to taxonomy. It was because `taxon_list.csv` is untracked so the taxonomy changes carried over from another branch, but I should have just reset it.

ehwenk (approved):
Looks perfect!

PR #132, opened 2023-11-16
More dataset tests

### LINKED ISSUE #118: Add `dataset_test` for checking duplicate rows in contexts table
Currently in AusInverTraits there are some context `value`'s that have different `description`'s. 

For example in Adams_1992: 
image

The value "pollinator" has been entered as a context `value` twice with two different `description`'s, leading to the same set of `link_vals`, meaning the contexts table joined onto the traits table will result in a warning of non-uniquely identified rows. 

We should add a `dataset_test` test to check for this issue.

yangsophieee:
- Fixes #118 
- Rename `check_duplicates` to `check_pivot_duplicates`
- Add option in `metadata_check_custom_R_code` `path_data` to allow easier testing in tests/testthat
- Add Test Datasets 5 and 6
- Update taxon_list-orig.csv
- Clean existing test metadata files


ehwenk (approved):
I've looked over the changes to the R files carefully - but can't be certain I've checked everything in the tests

PR #139, opened 2023-11-23
Update `taxon_list.csv` in tests

yangsophieee:
Allow R files to be tracked and add `build_update_taxon_list.R`

- I removed the "R/" from the .gitignore so that I could track `build_update_taxon_list.R` changes (it could potentially change in the future)
- Updated `taxon_list-orig.csv` in tests and expected outputs and `test-setup.R` accordingly


tests/testthat/examples/Test_2023_1/output/taxonomic_updates.csv
```diff
@@ -1,48 +1,48 @@
...
-Test_2023_1,Psychotria sp Utchee Creek,Psychotria sp. Utchee Creek (H.Flecker NQNC5313),species,Psychotria sp. Utchee Creek (H.Flecker NQNC5313),accepted,https://id.biodiversity.org.au/taxon/apni/51446727
-Test_2023_1,Quassia baileyana,Quassia baileyana,,Samadera baileyana,nomenclatural synonym,https://id.biodiversity.org.au/instance/apni/897542
-Test_2023_1,Rhodomyrtus trineura,Rhodomyrtus trineura,,Rhodomyrtus trineura,accepted,https://id.biodiversity.org.au/taxon/apni/51441176
-Test_2023_1,Rockinghamia angustifolia,Rockinghamia angustifolia,,Rockinghamia angustifolia,accepted,https://id.biodiversity.org.au/node/apni/2921179
-Test_2023_1,Syzygium,Syzygium,,,,
-Test_2023_1,Syzygium sayeri,Syzygium sayeri,,Syzygium sayeri,accepted,https://id.biodiversity.org.au/node/apni/2888968
-Test_2023_1,Testupdated sp4,Testupdated sp4 replaced,species,,,
-Test_2023_1,Trema aspera,Trema aspera,,Trema tomentosa var. aspera,nomenclatural synonym,https://id.biodiversity.org.au/instance/apni/856997
-Test_2023_1,testupdated sp3,testupdated sp3 replaced,species,,,
+dataset_id,original_name,aligned_name,taxonomic_resolution,taxon_name,aligned_name_taxon_id,aligned_name_taxonomic_status
```
ehwenk, line 1: For all of these, is the only change that `aligned_name_taxon_id` and `aligned_name_taxonomic_status` columns swapped order? And if so, do we want that? I'm actually inclined to change the code that selects what columns to maintain and keep the original order in these files. Because I've been tending to keep the identifiers at the far right of tables

yangsophieee: Yes I think that's the only change. Will have to re-update `build_update_taxon_list()` in austraits.build as well.

yangsophieee: Nevermind it would have to be in [line 1917](https://github.com/traitecoevo/traits.build/blob/93cfd3ef1cb57df7e93bc4d844128a0f6ffb5e49/R/process.R#L1917-L1918) in process.R - would need to list the `align` columns by name instead of `contains`. But we don't want to change the output structure again so will leave it as is.

Issue #141, opened 2023-11-23
Add test for checking units in metadata are correct/in the `unit_conversions` file


yangsophieee:
For `dataset_test`:
Surprised this doesn't exist right now. If a unit is not recognised in the `unit_conversions` file, the values go straight to `excluded_data` with no notification to the user. I can't think of a reason to not have this as a test.

Thanks @dcol2804 for bringing to my attention!

yangsophieee:
Implemented on branch `add_dataset_tests`

PR #143, opened 2023-11-24
Add dataset tests

yangsophieee:
Fixes https://github.com/traitecoevo/traits.build/issues/141
- Check units are found in the unit conversions file (I know this is addressed by the new `dataset_check` functions but I thought this would be good as a dataset test, not sure what you think @ehwenk)?
- Also added checks for duplicate trait `var_in` and duplicate context `var_in` and `context_property`


R/testdata.R
```diff
@@ -590,6 +590,27 @@ dataset_test_worker <-
             process_format_contexts(dataset_id, data)
         )
 
+        # Check that there are no duplicate `var_in` or `context_property` fields
+        context_properties <- sapply(metadata[["contexts"]], "[[", "context_property")
+        context_vars_in <- sapply(metadata[["contexts"]], "[[", "var_in")
+
+        expect_equal(
+          context_properties |> duplicated() |> sum(),
```
ehwenk, line 598: I feel like Daniel at some point asked that we just use `%>%` to standardise but I might be wrong

dfalster: Either is fine for now. At some point we might unify, but not everything is perfectly interchangeable so can't just find replace.

R/process.R
```diff
@@ -2050,7 +2050,7 @@ check_pivot_duplicates <- function(
 
   # Check for duplicates
   database_object$traits %>%
-    filter(.data$dataset_id %in% dataset_ids) %>%
+    dplyr::filter(.data$dataset_id %in% dataset_ids) %>%
     select(
```
ehwenk, line 2054: `dplyr::` here too

R/setup.R
```diff
@@ -900,15 +900,15 @@ metadata_add_taxonomic_change <- function(dataset_id, find, replace, reason, tax
...
+          dplyr::bind_rows(to_add) %>%
+          dplyr::filter(!.data$find == replace) %>%
+          arrange(.data$find)
       }
     } else {
       data <- dplyr::bind_rows(data, to_add) %>%
-            filter(!.data$find == replace) %>%
-            arrange(.data$find)
+        dplyr::filter(!.data$find == replace) %>%
+        arrange(.data$find)
```
ehwenk, line 911: `arrange` should also have `dplyr::` throughout (`arrange` also in `plyr` - and I think different functionality

R/testdata.R
```diff
@@ -1083,10 +1129,10 @@ dataset_test_worker <-
 
         ## Check traits are not only NAs
```
ehwenk, line 1130: Maybe better `Check that not all trait names are NAs`. When I read header I thought this was about trait values until I looked at code.

ehwenk:
> * Check units are found in the unit conversions file (I know this is address by the new `dataset_check` functions but I thought this would be good as a dataset test, not sure what you think @ehwenk)?

@yangsophieee I think it is fine as a `dataset_check` function. I do know that there has been the occasional study where it took us a long time (after datasets were merged into develop) to properly work out units for the occasional trait. Adding this as a test, effectively means `Missing unit conversion` will never actually be a reason for being in `excluded data` because tests won't allow that. But let's leave it for now and if it becomes a problem we can comment it out.

ehwenk (approved):
Thanks for making the requested changes. Can you put the note about the possible problem with the units test in the commit message, so we see it quickly in the future?

PR #144, opened 2023-12-03
Combined table output

ehwenk:
A function to generate a combined table output - trait data are still in long format, with all information from relational tables merged as additional, condensed columns.

Requires review and plan for how this will be integrated into workflow. For now I'm assuming it will be a function that can be sourced and run to generate outputs for releases.

@yangsophieee @dfalster It would be great if both of you could build the combined table and look at syntax, packing "rules".


R/output_combined_table.R
```diff
@@ -0,0 +1,137 @@
...
+
+  contexts_tmp <-
+    database$contexts %>%
+      dplyr::mutate(
+        context_property = stringr::str_replace_all(context_property, "=", "-"),
+        value = stringr::str_replace_all(value, "=", "-"),
+        description = stringr::str_replace_all(description, "=", "-"),
+        context_property = stringr::str_replace_all(context_property, ";", ","),
+        value = stringr::str_replace_all(value, ";", ","),
+        description = stringr::str_replace_all(description, "=", "-"),
```
yangsophieee: I think this should be ";", and ","

R/output_combined_table.R
```diff
@@ -0,0 +1,137 @@
...
+        value = ifelse(
+          is.na(description),
+          paste0(context_property, ":", value),
```
yangsophieee, line 78: Similarly here I wonder if "=" would be more readable than ":"

R/output_combined_table.R
```diff
@@ -0,0 +1,137 @@
...
+      ) %>%
+    dplyr::select(-dplyr::all_of(c("description", "context_property", "category"))) %>%
+    tidyr::separate_longer_delim(link_vals, ", ") %>% distinct()
+
+  reformat_contexts <- function(contexts_table, context_id) {
+    context_category <- gsub("_id", "_properties", context_id, fixed = TRUE)
+    out <- contexts_table %>%
+      dplyr::filter(link_id == context_id) %>%
+      dplyr::select(-link_id) %>%
+      dplyr::distinct(dataset_id, link_vals, .keep_all = TRUE)
```
yangsophieee, line 90: Is this line necessary? Shouldn't it be distinct anyway, otherwise there's something wrong with the contexts table that we want to be picked up by `dataset_test`?

R/output_combined_table.R
```diff
@@ -0,0 +1,135 @@
...
+      by = c("dataset_id", "temporal_context_id"),
+      reformat_contexts(contexts_tmp, "temporal_context_id")
+    ) %>%
+    dplyr::left_join(
+      by = c("dataset_id", "method_context_id"),
+      reformat_contexts(contexts_tmp, "method_context_id")
+    )
+  }
+
+  combined_table <-
```
yangsophieee, line 120: I'm wondering if the ID columns that help join to other relational tables could be removed from the combined table?

R/output_combined_table.R
yangsophieee: I've added a fix in a commit

yangsophieee (commented):
Still having a look at the final output, but submitting some comments for now.

R/output_combined_table.R
```diff
@@ -0,0 +1,134 @@
...
+      dplyr::ungroup() %>%
+      dplyr::distinct()
+
+  contexts_tmp <-
+    database$contexts %>%
+      dplyr::mutate(
+        context_property = stringr::str_replace_all(context_property, "=", "-"),
+        value = stringr::str_replace_all(value, "=", "-"),
+        description = stringr::str_replace_all(description, "=", "-"),
+        context_property = stringr::str_replace_all(context_property, ";", ","),
```
yangsophieee, line 73: Do the characters "<" and ">" also need to be replaced in case they ever make it into the `context_property`, `value` and `description` fields?

yangsophieee:
I've had a look at the final table. I think it looks good overall!

dfalster:
I will review the table next week. In the mean time, I agree we should move a copy of `Join_contexts` into traits.build

R/output_combined_table.R
```diff
@@ -0,0 +1,137 @@
...
+      dplyr::mutate(
+        location_property = stringr::str_replace_all(location_property, "=", "-"),
+        value = stringr::str_replace_all(value, "=", "-"),
+        location_property = stringr::str_replace_all(location_property, ";", ","),
+        value = stringr::str_replace_all(value, ";", ",")
+      ) %>%
+      dplyr::mutate(location_properties = paste0(location_property, "=", value)) %>%
+      dplyr::select(dplyr::all_of(c("dataset_id", "location_id", "location_name", "location_properties"))) %>%
+      dplyr::group_by(dataset_id, location_id, location_name) %>%
+      dplyr::mutate(location_properties = paste0(location_properties, collapse = "; ")) %>%
```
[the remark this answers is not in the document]
ehwenk: possibly but, I think `distinct()` is actually clearer in this case - otherwise you'd just have to list all the column individually and decide how to treat each one (lots of "take first value")

ehwenk:
This is now considered a stale branch and will shortly be deleted. The work on this function is now occurring at https://github.com/traitecoevo/austraits/tree/combined_table-functions. The core function, `database_create_combined_table` will be re-exported through traits.build once the next version of austraits is released.

PR #145, opened 2023-12-07
Move duplicate functions

### LINKED ISSUE #131: [traits.build tests] Duplicated tests between `dataset_test` and `tests`
Lots of tests duplicated between `dataset_test` and `tests` - need to review

yangsophieee:
- Fixes #131 
- Move custom `expect_` functions out of `dataset_test`, into R/test_functions.R so that R CMD check warnings are removed (otherwise "no visible global function definition" warning comes up)
- Rename prefixes to `test_expect_` to differentiate between custom `testthat` functions and built-in `testthat` functions
- I kept using `expect_silent()` from {testthat} instead of writing a custom function because I'm not sure how to rewrite the function to capture messages on top of errors and warnings


tests/testthat/test-setup.R
```diff
@@ -74,8 +74,7 @@ test_that("`metadata_create_template` is working with simulated user input", {
   expect_no_error(
```
dfalster, line 74: should be `test_no_error`?

tests/testthat/test-setup.R
```diff
@@ -720,7 +718,7 @@ test_that("`build_setup_pipeline` is working", {
 testthat::test_that("`dataset_find_taxon` is working", {
   expect_silent(suppressMessages(austraits <- remake::make("test_name")))
   taxon <- c("Acacia celsa", "Acronychia acidula", "Aleurites rockinghamensis", "Syzygium sayeri")
-  expect_no_error(x <- dataset_find_taxon(taxon, austraits), label = "`dataset_find_taxon`")
+  expect_no_error(x <- dataset_find_taxon(taxon, austraits))
```
dfalster, line 721: and here

tests/testthat/test-xamples.R
```diff
@@ -20,19 +20,17 @@ testthat::test_that("Test Dataset 1 builds correctly", {
...
-    info = "Building Test Dataset 1")
+    ))
 
   # Expected output
 
   tables <- c("traits", "locations", "contexts", "methods", "excluded_data",
               "taxonomic_updates", "taxa", "contributors")
 
-    expect_no_error(
+  expect_no_error(
```
dfalster, line 30: and here

tests/testthat/test-xamples.R
```diff
@@ -56,17 +54,15 @@ testthat::test_that("Test Dataset 2 builds correctly", {
...
       file.path(examples_dir, "Test_2023_2/data.csv"),
       "Test Dataset 2", definitions, unit_conversions, schema, resource_metadata, taxon_list
-    ),
-    info = "Building Test Dataset 2")
+    ))
 
   # Expected output
   tables <- c("traits", "locations", "contexts", "methods", "excluded_data",
               "taxonomic_updates", "taxa", "contributors")
   expect_no_error(
```
dfalster, line 62: and here

dfalster: Hmm, there a few places where this should be changed, so I won't mark everyone

R/test_functions.R
```diff
@@ -0,0 +1,377 @@
+
+test_no_error <- function(object, info) {
```
dfalster, line 2: We've dropped the `expect`. For all there's, I'd prefer `test_expect_...`

tests/testthat/test-xamples.R
```diff
@@ -289,17 +275,15 @@ testthat::test_that("Test Dataset 8 builds correctly", {
...
     expected_output <-
       purrr::map(
-        tables, ~read_csv(sprintf("examples/Test_2023_8/output/%s.csv", .x), col_types = cols(.default = "c"))),
-    info = "Reading in expected output tables"
```
dfalster, line 302: Why delete? The info isn't helpful?

tests/testthat/test-xamples.R
```diff
@@ -255,17 +243,15 @@ testthat::test_that("Test Dataset 7 builds correctly", {
...
+    ))
 
   # Expected output
   tables <- c("traits", "locations", "contexts", "methods", "excluded_data",
               "taxonomic_updates", "taxa", "contributors")
   expect_no_error(
     expected_output <-
       purrr::map(
-        tables, ~read_csv(sprintf("examples/Test_2023_7/output/%s.csv", .x), col_types = cols(.default = "c"))),
-    info = "Reading in expected output tables"
```
dfalster, line 268: Why delete? The info isn't helpful?

dfalster (changes requested):

R/test_functions.R
yangsophieee: Changed all `test_` to `test_expect_...`

tests/testthat/test-setup.R
yangsophieee: @dfalster The custom `test_expect_` functions in `test_functions.R` are only used in `dataset_test` where customisation of error messages is needed. The rest of the tests (`test-setup.R`, `test-xamples.R`, etc.) use testthat:: functions -- which is why I renamed all the custom test functions to `test_` instead of keeping the `expect_` syntax to avoid confusion.

tests/testthat/test-xamples.R
yangsophieee: It's using `expect_no_error` from {testthat} which does not accept an info argument.

dfalster: Ok, but you also had a new function defined called `expect_no_error`, which would overwrite the one from {test that}. Now it's renamed to `test_ expect_no_error`, so perhaps isn't being used?

dfalster:
> @dfalster I should have explained clearer but basically `R/test_functions.R` contains 'custom' `expect_` functions, now named `test_expect_` functions, which allow you to customise the failure message. The inbuilt `expect_` functions from {testthat} do not allow full customisation of failure messages. I have used the `test_expect_` functions only in `dataset_test()`, and the rest of the tests use {testthat} inbuilt functions. Hope that makes sense!

Ok!

tests/testthat/test-xamples.R
yangsophieee: Oh there shouldn't have been a new function defined called `expect_no_error`; if there was- that's my bad! 

I can't seem to find it though. I did a find and replace on `test_` to `test_expect_` before so theoretically I shouldn't have touched `expect_no_error`.

R/test_functions.R
```diff
@@ -0,0 +1,377 @@
+
+test_expect_no_error <- function(object, info) {
```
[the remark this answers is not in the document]
dfalster: This is not really a problem as such, so let's consider this resolved.

dfalster (approved):
@ehwenk good to merge !

PR #151, opened 2024-02-11
Bugfixes: Fix minor errors

dfalster:
- dates keep swapping between UK and US version each time the ontology table is opened. I'm now formatting with the actual month, to be more transparent
- links for individual terms weren't working
- change name for ontology landing to "index.html"
- regenerate files and docs

---------
    
Co-authored-by: Elizabeth Wenk <ehwenk@gmail.com>


dfalster:
Hi @ehwenk, this should fix the issues you raised in 3a3cc403775419434c57cf769e0e5ed353f1abf3 and also the name of file. Can you open the site from docs/index.html and check everything is now looking good?

ehwenk (approved):
I've confirmed that the index.html file in docs/ontology is ordered correctly and doesn't include any duplicate terms.

PR #177, opened 2024-11-19
Import austraits

ehwenk:
Removal of functions from traits.build to accompany release of austraits 3.0
- `util_df_to_list` and `util_list_to_df` are now sourced from `austraits` (as `convert_df_to_list` and `convert_list_to_df`)
- `build_combine` is now replaced by austraits::bind_databases
- `plot_trait_distribution_beeswarm`, `trait_pivot_longer` and `trait_pivot_wider` had been in both packages and have now been removed from traits.build
- the function `flatten_database` is a replacement for `database_create_combined_table`, which was still on a branch
- at the bottom of `utils.R`, the functions that have been moved and renamed are also assigned their old names, to ensure that the names included in our Eco Informatics publication are still "valid" and that people using old remake/build.R files are not inconvenienced by these changes


R/utils.R
```diff
@@ -320,3 +267,12 @@ create_tree_branch <- function(x, title, prefix = "") {
     )
   )
 }
+
+# Renaming of austraits functions to ensure old scripts still work
```
dfalster, line 271: nice!

R/utils.R
```diff
@@ -320,3 +267,12 @@ create_tree_branch <- function(x, title, prefix = "") {
...
+
+util_list_to_df1 <- austraits::convert_list_to_df1
```
dfalster, line 273: There might be a way to do this that also enables the help for these functions, leveraging the new name. 

I also wonder if we should flag these as deprecated? 

@fontikar What's your advice?

tests/testthat/test-setup.R
```diff
@@ -737,7 +737,7 @@ test_that("reports and plots are produced", {
   # Not testing right now
   #expect_no_error(
     #p <-
-      #traits.build::plot_trait_distribution_beeswarm(
+      #austraits::plot_trait_distribution_beeswarm(
```
dfalster, line 740: We should just delete these commented lines, as the function is tested in `austraits1

dfalster (changes requested):
Nice work! See query regarding deprecated functions.

R/utils.R
ehwenk: Yes, @fontikar and I talked about this, but then ran out of time on Friday. I agree it would be good to print a message when *some* of these are used.

But I think for `database_create_combined_table` we really should just have this name preserved for a while, without any message, since it is shown in the main figure in the publication under that name.

fontikar: Sorry team, for the late replies! 

Best practices are detailed [here ](https://lifecycle.r-lib.org/articles/communicate.html)

I think we need to bring back the old function for now so we can properly implement deprecation. such as `deprecate_soft()` or `deprecate_warn()`

Like this: 
```
add_two <- function(x, y) {
  lifecycle::[deprecate_warn](https://lifecycle.r-lib.org/reference/deprecate_soft.html)("1.0.0", "add_two()", "base::sum()")
  x + y
}
````
Alternatively, we can use `deprecate_stop()` function, forcing the using to update their code to the new function. I am in favour of this given as @dfalster the users of traits.build functions is likely still small-ish where you can be a bit more firm with the deprecation. 

I suggest we follow what is suggested here in [renaming a function](https://lifecycle.r-lib.org/articles/communicate.html#rename-a-function) 

thoughts?

Comment (fontikar) [reply R/utils.R:273]: I think we should also rexport so we can point to austrait's documentation

```
#' @importFrom magrittr %>%
#' @export
magrittr::`%>%`
```

dfalster: Thanks @fontikar I like your suggestion and commits.

dfalster:
Good start @ehwenk , good additions @fontikar . Just need to fix tests.

dfalster (approved):

PR #187, opened 2024-12-19
Add standard deviation to allowed value types

ehwenk:
Although SE mostly reported today, older manuscripts often document SD instead


dfalster (changes requested):
ah, needs a test!

dfalster (approved):

PR #206, opened 2025-06-12
Various small fixes

ehwenk:
Small fixes:
- regarding identifiers
1. early in coding the identifiers table I mistakenly added lines to read_metadata that created an element `identifier` instead of `identifiers`.
2. ensure that an empty identifiers item is added to the correct location
3. fix tests

- metadata_add_locations:
1. remove quotes around "location_name"; this was only removing the
"location_name" variable when it was actually "location_name" and
otherwise automatically adding it as another location_property

- dataset_report:
1. need to add "raw" values to value types that are plotted
2. ensure partial_matches_allowed set to "F" so if there are multiple
author_year dataset_id's only the intended one is extracted
3. additional report changes


inst/support/report_dataset.Rmd
```diff
@@ -509,31 +509,31 @@ writeLines(ifelse(length(traits_numeric) == 0,
         "No numerical traits data are currently available in this dataset",
         "We have recorded the following traits with continuous numerical values in this dataset. Plots are shown comparing the distribution of values in your study (blue) to those in other datasets within AusTraits (red, green). For each trait, please eyeball the information provided and let us know of any issues or updates where appropriate."))
 
-for (trait in traits_numeric) {
+for (trait in seq_along(1:length(traits_numeric)) {
```
dfalster, line 512: Let's review this part together

dfalster (approved):
Overall looks good, but just want to clarify the flagged coding issue with you

PR #221, opened 2026-07-27
Generate ids independently of the session locale (#29)

### LINKED ISSUE #29: `test_build_dataset` generates different id's when run within a `test_that` chunk versus a single line
When you run the `test_that` chunk in test-xamples.R, it generates different context id's (e.g. entity_context_id, method_id, etc.) to running each line within the chunk one-by-one. 

I noticed that when I changed the schema file (removed `value_type` from methods table) it also changed the id numbers.

dfalster:
Closes #29.

The reported symptom was that `test_build_dataset` produced different ids inside a `test_that` block than when run line by line. That happens because `test_that` pins collation to C via `local_reproducible_output()`. **The underlying problem is larger: a build on a contributor’s machine produces a different database than a build on CI.**

## Cause

Ids were derived from `sort()` and `as.factor()`, both of which collate character vectors according to `LC_COLLATE`:

```
C            NA_2_5  <  female_10_6  <  female_3_3  <  male_1_10
en_AU.UTF-8  female_10_6  <  female_3_3  <  male_1_10  <  NA_2_5
```

`tidyr::unite()` renders `NA` as the literal string `"NA"`, so wherever a context has missing values the position of that key moves and every id after it shifts. Two sites were affected: `process_generate_id()` (`observation_id`, `location_id`) and `process_create_context_ids()` (the context ids).

This was hard to spot because `dplyr::arrange()` — the *visible* ordering in `process.R` — already uses C collation and is stable. Only the invisible `sort()` / `as.factor()` steps moved.

Confirmed by building the examples twice, changing only `LC_COLLATE`: four of nine datasets differed, and the committed `examples/*/output/*.csv` only matched under C.

## Fix

Sort with `method = "radix"`, which always collates in the C locale, via two documented internal helpers. `util_index_locale_independent()` replaces `as.integer(as.factor(x))` and reproduces its NA handling exactly.

**Existing ids are unchanged.** C is the order CI has always built in, so the committed expected outputs still hold — verified across all nine examples.

`util_separate_and_sort()` had the same defect, silently reordering values *within* multi-value cells by locale, and is fixed the same way.

## On sorting policy

Worth settling as a house rule, since this will recur:

- **Determinism and human-friendly ordering are in tension.** Locale collation is nicer to read (`de_Kauwe_2015` filed under D, case-insensitive) but is not reproducible. C collation is reproducible but puts all upper case before lower case.
- **For anything feeding ids or file output, determinism has to win.** An id is a stable public identifier; it must not depend on who ran the build. Presentation-time ordering belongs in display code, not in the build.
- **C is already the de facto convention here**, since `dplyr::arrange()` defaults to it — so this change makes the codebase internally consistent rather than introducing a new convention.
- If you ever do want human-friendly *and* deterministic, `stringr::str_sort(x, locale = "en")` gives that (explicit ICU locale, independent of the session). But adopting it would renumber existing ids, so it would need to be a deliberate breaking change tied to a major version — not something to slip in.
- A deeper fragility remains, worth a separate issue: `process_generate_id(sort = FALSE)` — the default, used for `method_id`, `individual_id`, `population_id`, `repeat_measurements_id` — assigns ids by *order of first appearance*. That is deterministic for identical input, but any change to row order or to the schema renumbers everything, which is the other half of what the reporter observed.

## Tests

New `tests/testthat/test-locale.R` builds **every** example dataset under both `en_US.UTF-8` and C and asserts the results are identical, so any locale-sensitive sort added to the build in future is caught wherever it sits — not just at the two sites fixed here. Plus unit tests pinning the helpers’ collation and NA behaviour. Verified the test fails before the fix.


dfalster:
CI failure fixed in d3fecde — it was my test, not the change under review.

`config/taxon_list.csv` is **not in the repository**. It is created as a side effect of `test-setup.R:615`, which copies `config/taxon_list-orig.csv` over it. Test files run alphabetically, so `test-locale.R` sorted *before* `test-setup.R` and the file didn't exist yet:

```
'config/taxon_list.csv' does not exist in current working directory:
'D:/a/traits.build/traits.build/check/traits.build.Rcheck/tests/testthat'
```

Now reads the committed `taxon_list-orig.csv` fixture directly, so the file doesn't care what has already run. Verified against a clean clone with `taxon_list.csv` absent: all 12 assertions pass with the fix, and still fail without it.

Worth knowing more generally: **any new test file that sorts before `test-setup.R` will hit this**, and `test-xamples.R` only works today because `s` < `x`. Filed as #224.

PR #222, opened 2026-07-27
Stop dataset_test aborting on multiple identifiers (#218)

### LINKED ISSUE #218: dataset_test() aborts with "the condition has length > 1" for datasets with multiple identifiers
Found while investigating #213.

`dataset_test()` aborts entirely — not just fails a check — for any dataset that declares more than one identifier. This is currently the case for the bundled example `Test_2023_1`, and it is why `tests/testthat/test-xamples.R` fails on `develop`.

## Cause

`R/testdata.R` guards the identifiers checks with:

```r
if (!is.null(metadata$identifiers)) {
  if (!is.na(metadata$identifiers)) {
```

`metadata$identifiers` is a *list* of identifiers, so `is.na()` returns one value per identifier. Under R >= 4.2 an `if` condition of length > 1 is an error, so the whole run stops:

```
── Error: Test_2023_1 ──
Error in `if (!is.na(metadata$identifiers)) { ... }`: the condition has length > 1
```

The correct guard is already used for this same field further down the same file (~line 713):

```r
if ("identifiers" %in% names(metadata) & !all(is.na(metadata[["identifiers"]])))
```

## Second, independent bug in the same path

Once the guard is fixed, `dataset_test()` still fails for every dataset with identifiers:

```
Error: object 'schema' not found
```

`process_format_identifiers()` refers to `schema` in its body, but its parameter is named `traits`, and the caller passes `data`. Both this and the guard need fixing for `dataset_test()` to run on these datasets. The signature fix is already on the `fix/rdevel-check-identifiers-citation` branch.

Since identifiers are the headline 2.1.0 feature, this seems worth prioritising.

dfalster:
Closes #218.

`dataset_test()` aborts entirely — not just fails a check — for any dataset declaring more than one identifier. `Test_2023_1` is such a dataset, which is why `tests/testthat/test-xamples.R` currently fails on `develop`.

## Cause

`metadata$identifiers` holds a *list* of identifiers, so `is.na()` on it returns one value per identifier. Guarding with

```r
if (!is.na(metadata$identifiers))
```

therefore raises `the condition has length > 1` under R >= 4.2, stopping the whole run:

```
── Error: Test_2023_1 ──
Error in `if (!is.na(metadata$identifiers)) { ... }`: the condition has length > 1
```

## Fix

Use the guard already applied to this same field further down the file:

```r
if ("identifiers" %in% names(metadata) & !all(is.na(metadata[["identifiers"]])))
```

and collapse the two redundant nested conditionals into one block.

## Needs the signature fix too

With this alone, `dataset_test()` no longer aborts but still reports `object 'schema' not found` for every dataset with identifiers, because `process_format_identifiers()` refers to `schema` in its body while its parameter is named `traits` and the caller passes `data`. That is fixed on `fix/rdevel-check-identifiers-citation`.

With **both** applied I confirmed `dataset_test("Test_2023_1")` runs to completion and reports only the intentional failure:

```
examples/Test_2023_1/metadata.yml	traits - `trait_name`'s should not contain: 'wrong_trait_name'
```

so `test-xamples.R` should go green once both land. I have kept them separate rather than duplicating your commit — happy to rebase onto that branch instead if you would prefer them to go in together.

## Tests

New `tests/testthat/test-dataset-test.R` asserts `Test_2023_1` really does declare more than one identifier (so the test cannot silently stop covering the case) and that the run does not emit `the condition has length > 1`. Verified it fails before the fix.

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -56,6 +56,7 @@ Suggests:
     ggbeeswarm,
     gridExtra,
     scales,
+    withr,
     zip,
     covr
 Remotes:
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,3 +1,7 @@
+# traits.build (development version)
+
+- `dataset_test()` no longer aborts with `the condition has length > 1` for a dataset that declares more than one identifier. `metadata$identifiers` is a list, so `is.na()` on it returns one value per identifier.
+
 # traits.build 2.1.0
 
 - Identifiers table added, allowing trait values to be linked to a specific identifiers in an herbarium, museum collection, GenBank, or an arboretum. If a data contributor has collected data on the same individual plants across multiple datasets, these can also be linked.
--- a/R/testdata.R
+++ b/R/testdata.R
@@ -223,25 +223,21 @@ dataset_test_worker <-
         )
 
         ## Identifiers
-        if (!is.null(metadata$identifiers)) {
-          if (!is.na(metadata$identifiers)) {
-            testthat::expect_silent(
-              identifiers <-
-                metadata$identifiers %>%
-                process_format_identifiers(dataset_id, data)
-            )
-          }
-        }
-        
-        
-        if (!is.null(metadata$identifiers)) {
-          if (!is.na(metadata$identifiers)) {
+        # A dataset may declare several identifiers, so `metadata$identifiers`
+        # is a list and `is.na()` on it returns one value per element
+        if ("identifiers" %in% names(metadata) & !all(is.na(metadata[["identifiers"]]))) {
+
+          testthat::expect_silent(
+            identifiers <-
+              metadata$identifiers %>%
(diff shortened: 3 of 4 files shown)


dfalster:
Pushed a follow-up. CI was green, but the test was passing for the wrong reason.

`dataset_test()` reads `config/taxon_list.csv`, which is not in the repository — it is created as a side effect of `test-setup.R:615`. Test files run alphabetically, so `test-dataset-test.R` runs first on a clean checkout and the file is absent. The resulting error surfaces inside `dataset_test`'s own reporter rather than as an outer failure, so my assertion still passed.

It now seeds the fixture from the committed original when absent and cleans up afterwards. Re-verified against a clean clone with `taxon_list.csv` absent: passes with the fix, still fails without it — so it is genuinely exercising the guard.

See #224 for the underlying fragility.

Issue #225, opened 2026-07-27
Epic: streamline traits.build — usable, documented, maintainable, stable


dfalster:
Umbrella issue for the streamlining review. Full plan:
[`planning/2026-07-27-streamlining-review.md`](https://github.com/traitecoevo/traits.build/blob/docs/streamlining-review/planning/2026-07-27-streamlining-review.md) (branch `docs/streamlining-review`).

**Goals:** easy to use · well documented · easily maintained · **stable** (601 datasets across three
downstream databases depend on this package).

---

## Why start with testing

The test suite does not cover the whole output structure, and the gap predicts the bugs we found.

`test-xamples.R` compares **8 tables**; each `examples/Test_2023_*/output/` holds **12 artefacts**.
Never asserted: `sources.bib`, `definitions.yml`, output `metadata.yml`, and `identifiers.csv` for
8 of 9 datasets. `Test_2023_1` carries a 209 KB `identifiers.csv` that nothing compares — and it is
the multi-identifier case that crashes `dataset_test()`.

That single gap explains why **three independent bugs shipped in the 2.1.0 headline feature**:

| Fault | Location |
|---|---|
| `institution_code = .na` — `.na` is undefined; `metadata_add_identifiers()` always throws | `R/setup.R:624` |
| `process_format_identifiers()` uses `schema` but its parameter is named `traits` | `R/process.R:1677` |
| `is.na()` on a list of >1 identifier aborts the whole run | `R/testdata.R:227,238` |

`R CMD check` was already naming two of them — CI is configured with `error-on: "error"`, so both
WARNINGs and all 4 NOTEs are ignored.

The examples also miss vocabulary the real databases use: **13 of 14 `taxonomic_resolution` ranks**
(examples only ever use `species`), plus values of `basis_of_record`, `life_stage`, `entity_type`,
`value_type`. That matters because `R/process.R:1966` hardcodes a **botanical** rank list, and
`ausinvertraits.build` uses invertebrate ranks that aren't in it.

---

## Stages

- [x] **Stage 0 — close the coverage gap.** Assert all 12 output artefacts; add the missing
  vocabulary values to existing examples; re-enable `Test_2023_9` as the `dataset_test` *failure*
  test it was built to be (its README says so). Migrate to `expect_snapshot_value()`.
- [x] **Stage 1 — fix what the net catches.** Land #219 #220 #221 #222 and PR the unPR'd
[...]

PR #229, opened 2026-07-27
Clear the two R CMD check WARNINGs (non-ASCII source, Rd usage mismatch)

dfalster:
Raises the outstanding half of the never-PR'd `fix/rdevel-check-identifiers-citation` branch.

That branch was 3 commits ahead of `develop` but 9 behind, so it could not be pushed as-is —
its identifiers fixes have since landed independently via #222 and #227, and merging it would
have reverted #219–#222. This branch is cut fresh from `develop` and carries only what is
still missing.

### What this fixes

Both R CMD check WARNINGs, which is what makes this worth landing early: per the review in
#225, CI is currently set to `error-on: '"error"'` specifically because these WARNINGs were
present, and that loosening is what let the `.na` crash ship unnoticed. Clearing them is the
precondition for turning the check back into a bug detector.

| Fix | Why |
|---|---|
| `\uXXXX` escapes for the `check_disallowed_chars` exceptions string (`R/test_functions.R`) | Non-ASCII in R source is a CRAN portability WARNING. Also more legible than a wall of accented glyphs. |
| Reword two em-dashes in the `util_sort_locale_independent` roxygen block (`R/utils.R`) | Same portability check. These arrived with #221 earlier today, so this stops the WARNING reappearing the moment the other one is fixed. |
| Drop the stray `@param user_responses` from `metadata_add_identifiers()` | The function has no such argument, which is an Rd `\usage` mismatch WARNING. Implementing the argument is separate work (Stage 1). |

Plus one determinism fix that travelled with the same branch: `bib_print()` now normalises
page-range dashes to a plain hyphen. Newer RefManageR renders page ranges with an en-dash, so
`source_primary_citation` differed by dash type between RefManageR versions, and therefore
between machines.

### Verification

- Full suite: **921 passing, 0 failures, 0 warnings** on R 4.6.0.
- `man/util_sort_locale_independent.Rd` regenerated with roxygen2 8.0.0 (matching
  `Config/roxygen2/version`); no churn in any other Rd.
- Non-ASCII sweep across all of `R/` and `man/` now comes back empty.

### Note on the source branch

`fix/rdevel-check-identifiers-citation` is now fully superseded and can be deleted once this
merges. It never existed on the remote.

Prerequisite for Stage 0 of #225, which touches the same files.

### DIFF
--- a/R/process.R
+++ b/R/process.R
@@ -1074,6 +1074,9 @@ bib_print <- function(bib, .opts = list(first.inits = TRUE, max.names = 1000, st
     gsub("  ", " ", .) %>%
     gsub("DOI:", " doi: ", ., fixed = TRUE) %>%
     gsub("URL:", " url: ", ., fixed = TRUE) %>%
+    # Normalise page-range dashes to a plain hyphen so output is stable across
+    # RefManageR versions (newer versions render page ranges with an en-dash)
+    gsub("(pp?\\. \\d+)[\u2013\u2014](\\d+)", "\\1-\\2", .) %>%
     ifelse(tolower(bib$bibtype) == "article",  gsub("In:", " ", .), .)
 }
 
--- a/R/process.R
+++ b/R/process.R
@@ -1065,6 +1065,13 @@ bib_print <- function(bib, .opts = list(first.inits = TRUE, max.names = 1000, st
   oldopts <- RefManageR::BibOptions(.opts)
   on.exit(RefManageR::BibOptions(oldopts))
 
+  # RefManageR quotes titles with `dQuote()`, which honours `useFancyQuotes`.
+  # That option defaults to TRUE, so a plain `Rscript` build renders titles with
+  # curly quotes while testthat (which forces it FALSE) renders straight ones.
+  # Pin it so citations do not depend on how the build was launched.
+  oldquotes <- options(useFancyQuotes = FALSE)
+  on.exit(options(oldquotes), add = TRUE)
+
   bib %>%
     format.BibEntry(.sort = FALSE) %>%
     # HACK: remove some of formatting introduced in line above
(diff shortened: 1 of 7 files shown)


dfalster:
Added a second commit: **citations no longer depend on `getOption("useFancyQuotes")`**.

This turned up while starting Stage 0 (adding the missing output-artefact assertions), and
it is the same class of defect as the two already in this PR — output that varies with the
environment the build ran in.

RefManageR quotes titles with `dQuote()`, which honours `useFancyQuotes`. That option
defaults to `TRUE`, but `test_that()` forces it to `FALSE`. So:

- under the test suite, citations render with **straight** quotes — which is what every
  committed `output/methods.csv` fixture contains;
- under `Rscript`, `targets`, `remake` or an interactive session, they render with **curly**
  quotes (U+201C/U+201D).

Which means the databases actually built from this package in production have had curly
quotes in `source_primary_citation`, `source_secondary_citation` and
`source_original_dataset_citation` all along, and the suite could not possibly have caught
it: it only ever builds under testthat. Six of the nine examples differ from their golden
`methods.csv` when built outside a `test_that()` block.

Fixed by pinning the option inside `bib_print()`, with a test that builds `Test_2023_5`
under both settings and compares in full. Verified it fails without the fix (2 failures) and
passes with it. Suite green: 0 failures, `test-locale.R` now 3 tests / 14 expectations.

Worth a look at whether the downstream databases need a rebuild for this — the curly quotes
will be sitting in their published citation fields.

### Also worth recording

The branch this PR replaces, `fix/rdevel-check-identifiers-citation`, was **not** unPR'd as
#225 records — it was PR'd as #217 and **merged into `master`, not `develop`**, on 8 July.
That is why its fixes never reached `develop`. `master` is otherwise strictly behind
`develop`; the only other things it carries are #214 (AGENTS.md/CLAUDE.md pointer) and #215
(add-to-project workflow), which are still missing from `develop` and probably want
forward-porting separately.

PR #230, opened 2026-07-27
refactor stage 0: expand tests to assert the whole output structure, and cover the validation layer

dfalster:
Stage 0 of #225 — close the test coverage gap, because everything else in the plan depends
on that net existing. Builds on #229 (merged).

## The gap

`test-xamples.R` built nine example datasets and compared **eight tables**, while each
`examples/Test_2023_*/output/` holds **twelve artefacts**. `identifiers.csv` was compared for
one dataset of nine; `sources.bib`, `definitions.yml` and the output `metadata.yml` for none.
`Test_2023_9` was commented out entirely.

All twelve are now compared for all nine datasets, and the eight near-identical 30-line
blocks collapse into one loop.

## What the gap was hiding

The review predicted the newly-compared fixtures would fail. They did, and none of it was
blessed:

**`Test_2023_1/output/identifiers.csv` was wrong, not just stale.** 3358 rows of which 700
were distinct, and the observation→specimen mapping disagreed with the build. Observation 235
is recorded as `Plant 66`; the verified `traits.csv` puts observation 235 on taxon *Syzygium*,
which is the `Plant 73` row of `data.csv`. The build is right. Regenerated.

**`Test_2023_2`'s identifier never resolved.** It declares `var_in: specimen` and its README
has always described that column — but `specimen` was never committed to `data.csv`, in any
revision. So the build produced an empty identifiers table, **silently**, because
`process_add_all_columns` supplies a missing `var_in` as an all-NA column. A typo in `var_in`
costs you the entire identifiers table with no error, no warning. The committed 506-row
fixture was unreproducible from committed inputs. Column restored (two rows blank, so the NA
filter is exercised); the original values were not recoverable.

**`write_plaintext()` never exported identifiers at all.** The public exporter had a hardcoded
table list of eight, missing `identifiers` — so every export silently dropped 2.1.0's headline
table, for the whole release. This is a fourth independent identifiers bug beyond the three in
#225. It now derives the list from the database, and has a test; it had none.

**`dataset_test` produced invalid UTF-8.** `colour_characters()` split on bytes and wrapped
ANSI escapes around individual bytes, so reporting a non-ASCII character emitted an escape
*inside* a multi-byte sequence. `dataset_test` on Test_2023_9 aborted with `input string 285
is invalid UTF-8` before reporting anything. Both it and `check_disallowed_chars()` now work
per character. **Which characters get flagged is deliberately unchanged** — that rule is
looser than it should be, but tightening it changes which real datasets pass, so it belongs
behind the downstream build gate, not in this PR.

## Validation layer

Nothing verified that a failing check actually fails. `R/test_functions.R` had 2 of 30 helpers
referenced by any test, and the nearest thing to a check was `expect_output(dataset_test(...))`
with no `regexp` — which passes on any output at all.

`Test_2023_9` is re-enabled as the failure-output fixture it was built to be three years ago.
Its report is pinned as a snapshot: **28 distinct messages** across the context, trait,
metadata and taxonomic-update checks. The progress counter, the console-width dashed rules and
the backtraces (whose line numbers move whenever `R/` is edited) are stripped first, so the
snapshot doesn't churn on unrelated changes.

Test_2023_1's deliberate `wrong_trait_name` is now asserted by message.

## Vocabulary: taxonomic_resolution per rank

The rank branch in `dataset_update_taxonomy()` is extracted as `util_name_to_match_to()`, a
pure function, so each rank can be tested. The examples only ever use
`taxonomic_resolution: species`, leaving 13 of 14 ranks unexercised.

The invertebrate ranks turn out to be **handled correctly** — all six are coarser than
species, so first-word matching is right. The tests pin that, so a rank *below* species added
without being added to `ranks_at_or_below_species` fails loudly instead of being silently
truncated to its genus. They also record that an unrecognised rank falls back to the genus
without complaint.

## Regeneration path

`regenerate-examples.R` is new. Until now the only way to update a fixture was to hand-edit it
towards the observed output — which is exactly how `identifiers.csv` came to hold values no
build had produced in years. It regenerates through `write_plaintext()`, so the fixtures
exercise the exporter too, and pins `LC_COLLATE`/`useFancyQuotes` to what `test_that()` sets.

**Most of the fixture diff is representation, not behaviour**: `write_plaintext` writes NA as
empty where the hand-made files wrote `NA`, and `read_csv` reads the two alike. The
behavioural changes are the two identifiers files above.

## On `expect_snapshot_value()` — recommending against, with evidence

#225 proposes migrating the golden-file comparisons on the grounds that
`expect_equal(built[[v]], expected[[v]])` "dumps hundreds of rows for a one-cell change". That
was true under edition 2. **Under edition 3, which is already active, `expect_equal` routes
through waldo** and a one-cell change in the 527×26 `traits` table produces a 15-line diff
naming the exact cell with context. Measured, not assumed.

Migrating would replace nine readable, git-diffable CSVs per dataset with an opaque serialized
blob, and lose their second role as worked example output that the READMEs describe. The other
benefit sought — one-command regeneration — is what `regenerate-examples.R` provides, and it
additionally covers the exporter, which `snapshot_accept()` would not.

Snapshots **are** used where they fit: the `dataset_test` report is variable-length console
text, which is exactly the case `expect_snapshot` is for. `_snaps/` existed but was empty; it
now holds that.

## Verification

Coverage: `test-xamples.R` 8 tests/433 expectations → 10/514. Suite 820 → **856 passing, 0
failures, 0 warnings**.

Rather than assume the new assertions work, each was checked by breaking what it covers:

| Deliberate break | Caught by |
|---|---|
| identifiers `distinct()` removed | `test-xamples.R` (Test_2023_1, Test_2023_2) |
| `variety` dropped from the rank vocabulary | `test-taxonomic-resolution.R` |
| disallowed-character check stops firing | `test-dataset-test.R` snapshot |
| citation rendering unpinned | `test-locale.R` |
| `write_plaintext` drops identifiers again | `test-process.R` |
| trait values truncated in `dataset_process` | all nine datasets |

**Fresh clone** (`git clone` of this branch, no generated `taxon_list.csv`): 856 passing, 0
failures — identical to the working tree.

## Follow-ups this surfaced, filed separately

Each of these is a behaviour change rather than a coverage change, so per #225's own
doctrine they belong behind the downstream build gate rather than inside a test PR:

- **#232** — a `var_in` naming a column that doesn't exist silently yields an empty
  identifiers table. `process_add_all_columns` supplies the missing column as all-NA and the
  NA filter drops every row. This is why Test_2023_2's identifiers were empty for the
  dataset's entire lifetime.
- **#233** — `check_disallowed_chars` decides byte-wise, so a character assembled from bytes
  that all happen to occur in the exceptions string slips through. This PR fixed the
  *rendering* half (invalid UTF-8) but deliberately left detection alone, since tightening it
  can newly fail real datasets.
- **#234** — `dataset_test_worker` calls `local_edition(2)` (`R/testdata.R:60`). The package
  is on edition 3; the validation layer is not. That is the real edition-3 gap, and why the
  deprecated `context()` calls survive. The snapshot added here makes attempting the move
  materially safer.
- **#235** — Test_2023_9's metadata predates the `value_type`/`basis_of_value` and
  `taxonomic_resolution` schema entries, so 13 of the 28 snapshot messages are staleness
  rather than deliberate faults.
- **#236** — published databases carry curly quotes in their citation fields, a consequence
  of the `useFancyQuotes` bug fixed in #229. Needs a maintainer decision on whether to
  re-issue or fix on next build.

`master` carrying #214 and #215 is handled by **#231**.


dfalster:
### Fresh-clone verification (final branch)

`git clone` → `R CMD build` → `R CMD check --no-manual` on `190f15d`:

```
Status: 3 NOTEs
  * checking DESCRIPTION meta-information ... NOTE   (licence stub, invalid DCF)
  * checking dependencies in R code ... NOTE
  * checking R code for possible problems ... NOTE   (global variable bindings)
  Running 'testthat.R' ... OK
```

**0 errors, 0 warnings.** That is #225's Stage 2 target for warnings reached, which means
`error-on: "warning"` in `R-CMD-check.yml` can now be turned on — it was loosened to
`'"error"'` precisely because the WARNINGs were there, and that loosening is what let the
`.na` crash ship.

Getting there needed one more fix than #225 records. The review lists two Rd `\usage`
mismatches; there were three. `process_parse_data()` gained an `identifiers` parameter when
the identifiers feature landed and its roxygen was never updated, so the WARNING survived
#229. Fixed in the last commit.

Worth noting the suite also passes under `R CMD check` (i.e. against the built tarball in a
temp directory), not only under `test_local()` — so the new `_snaps/` directory and
`regenerate-examples.R` ship correctly and the snapshot resolves from the installed location.

The 3 NOTEs are all pre-existing and already tracked: the licence stub and `^ontology$` are
Stage 1, the global-variable bindings Stage 2.

PR #252, opened 2026-08-03
Add dataset_replace_disallowed_chars() to automate fixing special characters

### LINKED ISSUE #251: Automate replacing disallowed characters in metadata and data files
Follow-up to @ehwenk's review comment on #250:

> Looks good. We should also develop a workflow for automatically replacing disallowed characters, in data and metadata files

Fixing these by hand is slow, and #250 made the check stricter, so there is now more of it to do: **95 disallowed characters across 21 datasets** in the three downstream repos.

## The characters actually out there

Measured over every `metadata.yml` in `austraits.build` (411 datasets), `ausinvertraits.build` (160) and `AusFizz` (30), using the post-#250 check:

| character | count | reading |
|---|---|---|
| U+00BA `º` MASCULINE ORDINAL | 34 | `°` degree sign |
| U+223C `∼` TILDE OPERATOR | 17 | `~` approximately |
| U+2033 `″` DOUBLE PRIME | 9 | `"` arcseconds |
| U+00F1 `ñ` | 8 | legitimate (Briceño, El Niño) |
| U+2030 `‰` PER MILLE | 6 | legitimate symbol |
| U+0161 `š` | 6 | legitimate (Klimešová) |
| U+00D8 `Ø` | 5 | probably a mangled `Φ` in `ØPSII` |
| U+25E6 `◦` WHITE BULLET | 3 | `°` degree sign |
| U+00A0 NO-BREAK SPACE | 2 | a space |
| U+00AC `¬` + U+2020 `†` | 2 + 2 | see below |
| U+00D3 `Ó` | 1 | legitimate (Ósvaldsson) |

They fall into three groups, and only the first can be automated safely:

1. **Confusables** — a look-alike for a character that is already allowed, with the intent unambiguous from context (`33.33ºS`, `17ºC`, `16°17'24.8″S`, `∼38 Pa`). 69 of the 95.
2. **Mojibake** — `¬†` is a UTF-8 no-break space (bytes `C2 A0`) read back as Mac OS Roman. It is one artifact spanning two characters, so a per-character pass would mangle it.
3. **Letters and symbols** — either a real name or a mangling only the curator can interpret. These must be *reported*, never guessed at: silently rewriting `Briceño` or `Klimešová` would corrupt an author's name.

## Proposal

`dataset_replace_disallowed_chars(dataset_id, path_data = "data", files = c("metadata.yml", "data.csv"), dry_run = TRUE)`

- A curated map from confusable → intended character, plus the mojibake sequences. Longest key first, so `¬†` matches before its halves.
- Two invariants, asserted in tests: every key is currently *disallowed*, and every replacement value is *allowed*. So no entry is dead weight, and the output always passes `dataset_test()`.
- Anything not in the map is **reported with its code point**, not altered. Each is then either a character to add to `exceptions`, or a hand fix.
- Only the offending characters are rewritten — no round-trip through `write_metadata()`, which reflows the whole file and would bury the change in an unreviewable diff.
- `dry_run = TRUE` by default: see the report, then apply.
- `data.csv` is held to the ASCII-only standard `dataset_test()` already uses for data. Note the long-standing advice has been to fix data in `custom_R_code` rather than in `data.csv`, so this needs a maintainer call — see below.
- A file that is not valid UTF-8 is reported and skipped; recovering one needs the encoding it was written in, which the tool cannot know. (13 `data.csv` files are in this state.)

Expected effect on the backlog from #250: **69 of 95 auto-resolved, 26 reported for a decision** — and none of the 26 is a character the tool would have got right by guessing.

## Two things needing a maintainer decision

- **Should `data.csv` be rewritten in place?** `test_expect_allowed_text()` currently says *"Please replace using `custom_R_code`"*. Editing `data.csv` changes a file the contributor supplied. The alternative is to emit a `custom_R_code` snippet instead. Implemented as opt-in-by-dry-run for now.
- **`ñ`, `š`, `Ó`, `‰` are legitimate and recur.** They are flagged in 21 datasets today and every curator will hit them. Adding them to the default `exceptions` list would be a one-line change that removes 21 of the 26 remaining reports. Worth doing separately from this tool.

## Related, and probably the root cause

`traits.build-book`'s own guidance at [`adding_data_long.qmd:948`](https://github.com/traitecoevo/traits.build-book/blob/master/adding_data_long.qmd#L948) tells contributors to check that *"degree symbols (º)"* survive — and writes it with **U+00BA MASCULINE ORDINAL INDICATOR**, not U+00B0 DEGREE SIGN. That is the single most common disallowed character in the corpus, 34 of 95. The book is teaching the typo. Needs a fix in that repo.

dfalster:
Closes #251 — @ehwenk's follow-up to #250:

> Looks good. We should also develop a workflow for automatically replacing disallowed characters, in data and metadata files

Adds `dataset_replace_disallowed_chars()`. On the 95 disallowed characters that #250 surfaced across 21 datasets, it **resolves 69 automatically and reports 26 for a decision** — and none of the 26 is one it would have got right by guessing.

## What it does

```r
# See what would change, across every dataset. Writes nothing.
report <- dataset_replace_disallowed_chars(dir("data"))

# Apply it
dataset_replace_disallowed_chars(dir("data"), dry_run = FALSE)
```

```
44 character(s) can be replaced in 10 file(s) (U+00BA º x34, U+2033 ″ x9, U+25E6 ◦ x1)
25 character(s) have no known replacement: (U+00F1 ñ x8, U+0161 š x6, U+2030 ‰ x6, U+00D8 Ø x5)
	add each to `exceptions` if it is intended, or correct it by hand
Nothing written. Call again with `dry_run = FALSE` to apply.
```

## Three kinds of character, and only two are safe to automate

Measured over every `metadata.yml` in `austraits.build` (411 datasets), `ausinvertraits.build` (160) and `AusFizz` (30):

**Confusables** — a look-alike for a character already in the exception list, intent unambiguous from context. `º`→`°` (34, in `33.33ºS` and `17ºC`), `∼`→`~` (17), `″`→`"` (9, arcseconds in `16°17'24.8″S`), `◦`→`°` (3).

**Mojibake** — `¬†` is a UTF-8 no-break space (bytes `C2 A0`) read back as Mac OS Roman. It is *one artifact spanning two characters*, so keys are applied longest-first and it matches before its halves; a per-character pass would mangle it.

**Letters and symbols** — `ñ` (8), `š` (6), `‰` (6), `Ø` (5), `Ó` (1). Either a genuine name or a mangling only the curator can interpret, so these are **reported with their code point and never altered**. Silently rewriting `Briceño` or `Klimešová` would corrupt an author's name. `Ø` is probably a mangled `Φ` in `ØPSII`, which is exactly the kind of call a tool should not make.

## The map's two invariants

Asserted over **every** entry, so the map cannot rot:

- every key is currently **disallowed** — no entry is dead weight;
- every replacement value is **allowed** — the output always passes the check it exists to satisfy.

A separate test pins the property that matters most: the map **never touches a letter**, whether or not the check currently allows it.

## Diffs are surgical

This deliberately does *not* round-trip `metadata.yml` through `write_metadata()`, which reflows the whole file and would bury a one-character fix in an unreviewable diff. Verified by applying it to copies of four affected datasets (`Bourne_2017`, `Bryant_2021`, `Stanisic_2010`, `Ellsworth_2015`):

```
-    at Richmond, New South Wales, Australia (33.33ºS, 150.44ºE) near the Hawkesbury
+    at Richmond, New South Wales, Australia (33.33°S, 150.44°E) near the Hawkesbury
```

For all four: the YAML still parses, to the same structure, with **zero disallowed characters remaining**.

## Two things I'd like your call on

**1. `Stanisic_2010` has a duplicate hiding behind the mojibake.** Repairing it makes two `taxonomic_updates` entries byte-identical:

```
line 272  before: '- find: Coenocharopa\xa0yessabahensis\xa0'
line 282  before: '- find: Coenocharopa\xac†yessabahensis\xac†'
          after :  both '- find: Coenocharopa yessabahensis '
```

Same `replace:` value in both; the line-282 entry has `reason: .na.character` where line 272 has a real reason. So one should probably be deleted — but that is a curation decision in `ausinvertraits.build`, and I have not touched it. Note both also carry a **trailing space** in the taxon name.

**2. Should `data.csv` be rewritten in place?** `test_expect_allowed_text()` currently says *"Please replace using `custom_R_code`"*, and `data.csv` is a file the contributor supplied. It is included in `files` here but guarded by `dry_run`. Happy to drop it to metadata-only, or to have it emit a `custom_R_code` snippet instead — say which you'd prefer.

Separately, adding `ñ`, `š`, `Ó` and `‰` to the default `exceptions` would remove 21 of the 26 remaining reports in one line. I left that out as a judgement call for you rather than folding it in here.

## Also worth knowing

The book is teaching the most common typo. `traits.build-book`'s guidance at [`adding_data_long.qmd:948`](https://github.com/traitecoevo/traits.build-book/blob/master/adding_data_long.qmd#L948) tells contributors to check that *"degree symbols (º)"* survive — written with **U+00BA MASCULINE ORDINAL INDICATOR**, not U+00B0 DEGREE SIGN. That character is 34 of the 95 occurrences. Needs a fix in that repo; flagged in #251.

Files that are not valid UTF-8 (13 `data.csv` files) are reported and skipped — recovering one needs the encoding it was written in, which the tool cannot know.

## Tests

24 new tests in `test-replace-chars.R`, covering both map invariants, the multi-character mojibake ordering, every character measured in real data, letters being left alone, idempotence, code-point reporting (including a lone invalid byte reported as `0xB1` rather than a code point), the metadata/data strictness split, `dry_run` writing nothing, unresolvable characters surviving untouched, a resolvable and an unresolvable character on the same line, invalid UTF-8 being skipped byte-for-byte, and multi-dataset reporting.

Suite run with `NOT_CRAN=true`, since the `Test_2023_9` snapshot test is `skip_on_cran()`: **992 passing, 0 failures, 0 skips**, snapshot unchanged. Local CI-equivalent `R CMD check` reports `Status: OK`, including `checking code files for non-ASCII characters ... OK` — the map is written as `\uXXXX` escapes, since several of its entries are invisible or indistinguishable from their replacement in a source file.


dfalster:
Tidied after a review question about the number of functions: it was 9 for one feature, each internal one carrying a generated `man/*.Rd`. Three didn't earn the indirection and are now folded into their callers (`util_describe_char`, `util_tally_chars`, `util_empty_char_report`), leaving **five functions and one constant, and three fewer .Rd files**.

No behaviour change: 992 tests still passing, and the dry run over all 601 downstream datasets reports the same 69 replaceable and 26 unresolvable characters.

One tidy-up I started and backed out, in case it looks tempting later: reusing the new `util_write_lines_utf8()` inside `write_metadata()`. That function ends with `cat(txt, file = ...)`, and `writeLines()` appends a trailing newline — so it would have added a spurious blank line to every `metadata.yml` written by every `metadata_add_*` function. Not worth three lines.

dfalster:
@ehwenk — a question before I finish this, because the answer changes the design and I don't want to guess.

The intent is that `data.csv` gets **written in place** rather than fixed via `custom_R_code`. Happy to do that. But I measured what it would actually touch, and `data.csv` is a very different problem from `metadata.yml`:

| | `metadata.yml` | `data.csv` |
|---|---|---|
| characters flagged | **95** | **170,946** |
| files affected | 21 | **147** |
| would be rewritten | 69 | **125,657** |
| still flagged after | 26 | **43,669** |
| in files that aren't valid UTF-8 | 0 | 1,620 |

## Why the gap is so large

`data.csv` is held to ASCII-only (no exceptions), which is what `dataset_test()` uses for data. But `dataset_test()` only applies that to `parsed_data$traits$value` — the values that actually reach the built database. This tool reads **every line of the file**, so it sees every column, including ones that never become a trait value.

That difference is nearly all of the 170,946. The top two offenders:

```
RBGV_2023          110,666 x U+200D ZERO WIDTH JOINER   all in column 'created_by'
eFLOWER_Dun_2022     3,520 x U+2013 EN DASH             in 'Reference' and 'FullRef'
                       582 x U+00E9 é                   in 'User'
```

So the bulk is invisible junk in a provenance column, plus legitimate typography (en dashes in citation page ranges) and legitimate accented letters in people's names. **`dataset_test()` reports none of it today.**

Also worth knowing: `U+00B0 °` is flagged **6,858** times in `data.csv` — the degree sign is in the metadata exception list but data allows no exceptions. And there are 625 x `U+FFFD REPLACEMENT CHARACTER`, which is encoding damage that has already destroyed the original character; nothing can recover those.

## What I'd like you to pick

**A. `metadata.yml` only** — drop `data.csv` from the default `files`. Fixes the 95 that `dataset_test()` actually complains about; leaves `data.csv` alone.

**B. `data.csv` in place, whole file** — what you asked for, literally. Cleans 125,657 characters across 147 files, mostly deleting invisible ZWJs from `created_by`. Large diffs in contributor-supplied files, for problems nothing is currently reporting.

**C. `data.csv` in place, but only the columns that become trait values** — matches `dataset_test()`'s scope exactly, so the tool fixes precisely what the validator complains about and nothing else. More work: needs `process_parse_data()` to know which columns are in play.

**D. Emit a `custom_R_code` snippet** instead of touching the file, per the current message in `test_expect_allowed_text()`.

**My recommendation is C**, with B available via an explicit argument. C keeps "what the validator flags" and "what the fixer fixes" as the same set, which seems like the property you'd want long-term — otherwise `dataset_test()` stays quiet while the fixer wants to rewrite 147 files.

Whichever you pick, the 43,669 leftovers point at a separate question: **should `data.csv` really be ASCII-only?** En dashes in page ranges and accented letters in contributor names are not errors, and under the current rule they can never be resolved — only reported forever. Letting data share the metadata exception list would drop those 43,669 to roughly 128,000 → a few thousand real problems. That's your call, not mine, and I'd do it as its own PR.

The `metadata.yml` half of this PR is done and green either way, so I can also split `data.csv` out into a follow-up if you'd rather land the safe part now.

dfalster:
Pushed one defensive change while that question is open: **`data.csv` is no longer in the default `files`**.

The reason is narrow — as it stood, the documented one-liner `dataset_replace_disallowed_chars(dir("data"), dry_run = FALSE)` would have rewritten 147 `data.csv` files. The safe scope shouldn't depend on remembering to narrow it, so the default is now `metadata.yml` only and `data.csv` is processed on request.

Nothing is removed, and it's a one-line default to flip once you've picked from the options above. The trade-off is now documented under its own `@section` in the function's help, with the measured numbers, so whoever reads `?dataset_replace_disallowed_chars` next sees why the two files are treated differently.

994 tests passing.

ehwenk:
I think option C is good. It would help if the data.csv files came in clean.

> ## What I'd like you to pick
> **A. `metadata.yml` only** — drop `data.csv` from the default `files`. Fixes the 95 that `dataset_test()` actually complains about; leaves `data.csv` alone.
> 
> **B. `data.csv` in place, whole file** — what you asked for, literally. Cleans 125,657 characters across 147 files, mostly deleting invisible ZWJs from `created_by`. Large diffs in contributor-supplied files, for problems nothing is currently reporting.
> 
> **C. `data.csv` in place, but only the columns that become trait values** — matches `dataset_test()`'s scope exactly, so the tool fixes precisely what the validator complains about and nothing else. More work: needs `process_parse_data()` to know which columns are in play.
> 
> **D. Emit a `custom_R_code` snippet** instead of touching the file, per the current message in `test_expect_allowed_text()`.
> 
> **My recommendation is C**, with B available via an explicit argument. C keeps "what the validator flags" and "what the fixer fixes" as the same set, which seems like the property you'd want long-term — otherwise `dataset_test()` stays quiet while the fixer wants to rewrite 147 files.

ehwenk:
I'm confused as to what this new function does - does it run in the background during the build process? do you call it as part of adding/testing a dataset? 

I do feel that any "special characters" we can retain are good - they have a meaning, both in people's names and symbols with inherent technical meanings. I'm not clear what this function does with those.

dfalster:
@ehwenk Good questions, and your instinct on retaining characters was right — it caught a real flaw. I've changed the code. Taking your two questions in turn.

## When does it run?

**Nowhere automatically.** It is not in the build, not in `dataset_test()`, and nothing calls it. It is a command you run by hand when you want it:

```r
# after dataset_test() complains about characters, see what's fixable
dataset_replace_disallowed_chars("Bourne_2017")

# fix it
dataset_replace_disallowed_chars("Bourne_2017", dry_run = FALSE)
```

It edits `metadata.yml` on disk and that is all it does. Nothing in the build path changed, and no database needs rebuilding. `dry_run = TRUE` is the default, so the first call only ever prints a report.

## What it does with meaningful characters — and what you caught

It never touches letters, so names were always safe. But you were right that the map went further than it should have. Six entries were rewriting characters that **are not mistakes**:

| was doing | why that was wrong |
|---|---|
| `″` → `"` | U+2033 DOUBLE PRIME *is* the correct character for arcseconds |
| `′` → `'` | likewise for arcminutes |
| `∼` → `~` | TILDE OPERATOR carries "approximately" |
| `−` → `-` | MINUS SIGN is not a hyphen |
| `⁄` → `/` | FRACTION SLASH |
| `…` → `...` | ellipsis |

**All six are now removed.** Flattening them to ASCII loses information rather than fixing an error, which is the opposite of what the tool is for.

What is left is only characters that are genuinely *wrong* — 9 entries, down from 15:

- `º` MASCULINE ORDINAL and `◦` WHITE BULLET, both being used as a degree sign. These aren't degree signs at all; correcting them to `°` **recovers** the meaning rather than removing it.
- The mojibake pair `¬†`, which is a no-break space that got decoded as Mac OS Roman.
- Invisible characters with no content: no-break and zero-width spaces.

Everything else is reported with its code point and left alone. Across the three repos that changes the metadata split from 69 replaced / 26 reported to **43 replaced / 52 reported** — and the 52 are precisely the ones you'd want to decide about: `″` (9), `ñ` (8), `š` (6), `‰` (6), `Ø` (5), `∼` (17), `Ó` (1).

## Which brings me to the thing I think you actually want

If meaningful characters should be retained, the lever isn't this function — it's the **`exceptions` list**, which is what decides retention. Two findings there:

**For `metadata.yml`, the whole problem is 7 characters.** Adding `″ ∼ ñ ‰ š Ø Ó` to the default `exceptions` retains all 52 remaining occurrences, and then every single thing left flagged in metadata is a genuine error the tool can fix. That's a one-line change. (`Ø` is the one I'd flag — in `Brodribb_2000` it's `ØPSII`, which I suspect is a mangled `Φ`, so retaining it preserves a mistake. Your call.)

**For `data.csv`, it's much bigger, and it's the same cause.** Data is checked with `exceptions = ""` — ASCII only, no exceptions at all. So **42,463 characters that are already allowed in `metadata.yml` are rejected in `data.csv`**:

```
–  21,533   en dash        (already in the exception list)
°   6,858   degree sign    (already in the exception list)
é   3,432   
ü   2,856
—   1,904   em dash
ö     773      ó  746      ±  692      ’  638      ×  555      í  534
```

These are exactly the "symbols with inherent technical meaning" and accented names you're describing, and under the current rule they can never be resolved — only reported forever. Letting `data.csv` share the metadata exception list would retain all 42,463 of them.

I have **not** made either change, because the exception list is the vocabulary and that's yours to set. Say the word on the 7 metadata characters and I'll push it; the `data.csv` question probably wants its own PR since it interacts with the scope question above (whole file vs. only the columns that become trait values).

One last thing worth knowing regardless: 625 occurrences of `U+FFFD REPLACEMENT CHARACTER` in `data.csv` are encoding damage that has **already destroyed** the original character. Those can't be retained or repaired by anyone — the information is gone. Worth knowing they exist.

ehwenk (approved):
I think this should work well.

I've been replacing all the `U+2033 DOUBLE PRIME` and SINGLE PRIME with "straight quotes" since the beginning of AusTraits, but going forward we can leave them! But this will be inconsistent across metadata files.

dfalster:
Rebased onto `develop` (now including #253) and pushed two commits.

**`μ` U+03BC → `µ` U+00B5 added to the replacement map**, per @ehwenk's request on #253. Measured occurrences: 0 in any `metadata.yml` across all three databases, 2 in `ausinvertraits.build/data/Koster_1999/data.csv`, and 13 in `raw/` files that are never built. Since `data.csv` is out of the default scope, the entry only fires when `data.csv` is passed explicitly.

**Three tests were pinning behaviour #253 deliberately removed.** They asserted `″`, `‰` and `Ø` are still reported as disallowed; #253 makes all three allowed. Each PR passed CI alone, and the combination failed — the tests now use `†` and `∞`, which are disallowed both before and after #253 and are absent from the map, so they no longer move when the allowed list is widened. The arcseconds case now asserts `″` is left alone and not reported, which is what "leave correct typography in place" means under the new list.

Verified locally on develop + this branch: **1027 passing, 0 failures, 0 skips** with `NOT_CRAN=true`.

PR #253, opened 2026-08-03
Move the allowed-character list into the schema, and define letters by range

dfalster:
Refs #251, and answers two questions from @ehwenk's review of #252: where the exceptions live, and whether other surname letters are missing.

## Where they lived, and why that was the bug

In the **signature** of `check_disallowed_chars()` — a 343-character string literal — with a second, different rule (`exceptions = c("")`) hardcoded at its data call site. No single place to look, and editing it meant editing R code.

They now live in `inst/support/traits.build_schema.yml` under `allowed_characters`, read via `get_schema()`. One global definition, documented character by character, editable without touching R. Cached in a package-local environment, since the schema is a 500-line file and the check runs once per line of every `metadata.yml`.

## Yes — there were many missing letters, and the cause was systematic

The old list was **case-asymmetric**. 14 letters were allowed in lower case but not upper:

```
í č ó ö ø é è ł ń ê ù ú û µ
```

So `ó` passed and `Ó` failed. That is the entire reason `Ósvaldsson` was reported — not a policy decision about Icelandic names, just a hand-written list that missed the capitals. `Klimešová` failed for the same reason in reverse: `š` was never listed at all. The list also contained **two duplicate entries** (`í`, `µ`) and a **stray ASCII `l`**, which suggests it was assembled by hand in a hurry.

So rather than add the 16 letters I found and wait for the next name to break it, **letters are now defined by range**: Latin-1 Supplement and Latin Extended-A (`00C0`–`017F`). That covers the accented letters of every European language, and a letter can never again be treated differently from its own capital.

Two traps that the range approach has to handle, both now explicit in the schema:

- **Ranges contribute only their letters.** `00D7` MULTIPLICATION SIGN and `00F7` DIVISION SIGN sit inside the range but are not letters, so they aren't admitted by accident. `00D7` is listed deliberately under `symbols`; `00F7` is not allowed.
- **Unicode classifies the ordinal indicators as letters.** So the range *would* have admitted `00AA ª` and `00BA º` — and `º` is the single most common disallowed character in the database, being used for the degree sign. Admitting it would have silently undone the main fix in #250. Both are listed under `never_allowed`.

## Symbols that carry meaning are now retained

Per your point that symbols have inherent technical meaning: `2032` PRIME and `2033` DOUBLE PRIME (arcminutes/arcseconds), `2030` PER MILLE, `2212` MINUS SIGN, `223C` TILDE OPERATOR, `2044` FRACTION SLASH, `2026` ELLIPSIS.

## Measured effect: this is a strict loosening

The only two characters no longer listed are ASCII `l` and `-`, which are always allowed regardless — so **nothing that passed before can now fail.** Over every `metadata.yml` in all three repos:

| | before | after |
|---|---|---|
| disallowed characters | 95 | **43** |
| datasets affected | 21 | **8** |

And all 43 that remain are the five genuinely wrong characters — `º` (34), `◦` (3), no-break space (2), and the `¬†` mojibake pair (2+2) — every one of which `dataset_replace_disallowed_chars()` in #252 corrects automatically.

**Together the two PRs take the manual character-editing burden to zero:** the exception list retains everything meaningful, and the fixer corrects everything that is actually wrong.

## Deliberately not changed

The data call site still passes `exceptions = ""`, so `data.csv` stays ASCII-only. That's the open question on #251, where **42,463 characters already allowed in metadata are rejected in data** (21,533 en dashes, 6,858 degree signs, 3,432 `é`…). Worth doing, but it interacts with the whole-file-vs-trait-columns scope question, so it wants its own PR and your decision first.

Also left alone, and worth knowing about:

- **`Ø` is now allowed**, as a Danish/Norwegian letter. But in `Brodribb_2000` it appears as `ØPSII`, which I'm fairly confident is a mangled `Φ` (PSII quantum yield). Retaining it preserves a mistake — flagging it here rather than deciding.
- **`03BC μ` GREEK SMALL LETTER MU is still disallowed**, while `00B5 µ` MICRO SIGN is allowed. They look identical. Keeping only one means units stay consistent, but it will report 2 occurrences until someone normalises them. Say if you'd rather allow both.
- **CJK characters** (39 distinct, Chinese author names and one journal title) appear only in `data.csv`, so they're governed by the ASCII-only question above rather than by this PR.

## Tests

Four new tests: the allowed set is case-symmetric (asserting the only asymmetric members are the four whose other case is ASCII or Greek — `İ ı ſ µ`); the ordinal indicators stay disallowed despite being letters; the real names and symbols that used to be reported now pass; and the cache returns an identical value rather than re-reading.

`test-get.R` needed updating for the new top-level schema key — it pins the schema's shape, and correctly caught the change.

913 tests passing, 0 failures, 0 skips. CI-equivalent `R CMD check` reports `Status: OK`.


ehwenk:
`03BC μ GREEK SMALL LETTER MU is still disallowed, while 00B5 µ MICRO SIGN is allowed. They look identical. Keeping only one means units stay consistent, but it will report 2 occurrences until someone normalises them. Say if you'd rather allow both. `- Can this be added to the automated replacements?

`Φ` should be retained for its technical meaning. It is one of the symbols I've been changing manual to the textual "psi" in all datasets

inst/support/traits.build_schema.yml
```diff
@@ -475,3 +475,43 @@ metadata:
...
+  type: character_set
+  letter_ranges:
+    description: Letters are allowed by range rather than listed one at a time, so that a letter and its upper-case form can never be treated differently. Enumerating them by hand left 14 letters allowed in lower case but not upper, which is why a name like Osvaldsson was reported while osvaldsson would not have been. The range covers the accented letters of every European language, which is what contributor, author and taxon names need.
+    values:
+      00C0-017F: Latin-1 Supplement letters and Latin Extended-A. Code points in this range that are not letters are excluded automatically, so 00D7 MULTIPLICATION SIGN and 00F7 DIVISION SIGN are not admitted here; 00D7 is listed under symbols below.
+  letters:
+    description: Letters allowed from outside the ranges above.
+    values:
+      00B5: MICRO SIGN, as in micrometre. Note that 03BC GREEK SMALL LETTER MU is a different code point that looks identical; it is not allowed, so that units are written consistently.
```
ehwenk, line 489: MU will usually be a mis-presentation of the MICRO SIGN and an automated replacement should occur

Since both previously were disallowed lots of both symbols have been replaced with `u`

ehwenk (approved):
This is a good fix. Now that the default is to leave in characters and not to spend time searching out and manually replacing lots of characters, various "illegal" ones will slowly appear and can be added to the exclusion/inclusion lists.

inst/support/traits.build_schema.yml
dfalster: Agreed, and done — added in #252, which merges straight after this.

`μ` U+03BC → `µ` U+00B5 is now in `util_disallowed_char_replacements()`. It satisfies the map's invariant (disallowed on the left, allowed on the right), so the fixer will converge both spellings on the micro sign rather than reporting one and allowing the other.

On "lots of both symbols have been replaced with `u`" — I measured what is actually left, and the Greek letter is rarer than expected:

| where | GREEK MU U+03BC | MICRO SIGN U+00B5 |
|---|---|---|
| any `metadata.yml`, all 3 databases | **0** | 35 |
| `ausinvertraits.build/data/Koster_1999/data.csv` | 2 | – |
| `austraits.build/data/NHNSW_2023/raw/PlantNET_raw.csv` | 7 | – |
| `AusFizz/data/Aspinwall_2017/raw/notes.txt` | 6 | – |

So the two occurrences you saw are in a `data.csv`, and the other 13 are in `raw/` files that are never built. Since #252 keeps `data.csv` out of the default scope, the new entry will not fire until `data.csv` is passed explicitly — the Koster_1999 pair needs `files = "data.csv"`, or option C once that lands.

Separately, your point about `Φ` is still open: it is currently disallowed, so it gets reported rather than retained. I have not changed that here since it is a policy call — filing it as a follow-up.

PR #256, opened 2026-08-04
Finish the synonym replacement plumbing, and let curated substitutions win

### LINKED ISSUE #254: Tests indicate `definitions` is missing
The new traits.build tests are indicating the `definitions `are missing - but the dataset builds properly.
```
> dataset_test(current_study)
[ FAIL 2 | WARN 1 | SKIP 0 | PASS 84 ]

── Failure: Osmolovsky_2026 ─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
Osmolovsky_2026	`process_parse_data` threw an error:

Error in process_parse_data(., dataset_id, metadata, contexts, schema, : argument "definitions" is missing, with no default


── Warning: Osmolovsky_2026 ─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
restarting interrupted promise evaluation
Backtrace:
    ▆
 1. └─traits.build:::test_expect_no_error(...) at traits.build/R/testdata.R:723:9

── Error: Osmolovsky_2026 ───────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
<getvarError/missingArgError/error/condition>
Error in `process_parse_data(., dataset_id, metadata, contexts, schema, identifiers)`: argument "definitions" is missing, with no default
Backtrace:
    ▆
 1. ├─traits.build:::test_expect_no_error(...) at traits.build/R/testdata.R:723:9
 2. ├─data %>% ... at traits.build/R/test_functions.R:13:3
 3. └─traits.build:::process_parse_data(...)
[ FAIL 2 | WARN 1 | SKIP 0 | PASS 84 ]

dfalster:
Rebase of `adding-synonym-automated-replacements` onto `develop`, plus the fix for #254 and tests for the feature.

@ehwenk — the feature works; this finishes the plumbing and changes one thing you'll want to weigh in on (the ordering, below). Your two commits are preserved as-is; everything I changed is in the third commit. I opened this as a new branch rather than force-pushing yours, so `adding-synonym-automated-replacements` is untouched.

Closes #254.

## Why #254 happened

`process_parse_data()` gained a `definitions` argument, and `dataset_process()` was updated to pass it — but `dataset_test_worker()` was not. So every dataset built correctly and then failed its own test. It only ever reproduced on that branch; `develop` was never affected.

The reason it isn't a one-word fix: `definitions` means two different things here. `dataset_test()` holds the **whole** traits.yml object (`description`, `type`, `elements`), while the `process_*` functions expect the inner **`elements`** list keyed by `trait_name` (that's what `dataset_configure()` hands to `dataset_process()`). Passing the whole object doesn't degrade gracefully — it dies on the atomic `description` entry with `$ operator is invalid for atomic vectors`. So the call passes `definitions[["elements"]]`.

I deliberately did **not** give `definitions` a default. With `= NULL` a caller that forgets it would silently skip synonym replacement, so `dataset_test()` would validate data the build never produces. A loud error is the better failure. The existing `test-dataset-test.R` already covers this — I confirmed it fails with the exact #254 message when the fix is reverted, so CI would have caught it. The branch predated that test file, which is why it slipped through.

## The one behaviour change: substitutions now run before synonyms

This is the change worth your review. The synonym pass ran *before* dataset `substitutions`; it now runs *after*.

Because substitutions match on whole-value equality, a synonym that had already rewritten a value meant the curator's substitution could never match — silently, with no warning. Measured against all 411 `austraits.build` datasets:

| | count |
|---|---|
| Substitution entries silently pre-empted | **196**, in 21 datasets |
| ...where curator and synonym disagree outright | 5, in 5 datasets |
| Total substitution entries scanned | 3,990, in 193 datasets |

The disagreements are the clearest cases — the curator's considered value was being discarded in favour of the database-wide default:

| trait | value | curator says | synonym says |
|---|---|---|---|
| `plant_growth_form` | `vine` | `climber_woody` | `climber_herbaceous` |
| `bud_bank_location` | `rootstock` | `bud-bearing_root` / `basal_buds` | `basal_stem_buds` |
| `leaf_shape` | `cuneate` | `na` | `obtriangular` |

The other 191 are whole-value substitutions containing a synonym as a word — e.g. `dispersal_syndrome: "wind water"` was rewritten to `"anemochory hydrochory"` before the substitution matching `"wind water"` could fire.

Running substitutions first gives dataset-specific curation precedence over the database-wide default, and synonyms still resolve anything left over — including values a substitution *produces*. If you'd rather synonyms won, say so and I'll revert it, but the 196 silent no-ops look like the wrong default.

## Other changes

- **Extracted `process_replace_synonyms()`** so the pass is testable on its own, rather than being 50 lines inside an already-long function.
- **Replace one trait at a time** instead of joining a list column onto every row. 27 austraits traits declare synonyms while the long-format data runs to millions of rows, and the row-wise `map2_chr()` walked all of them: **2.7s vs 0.03s per 200k rows**. This also matches the `for`/`which` idiom the substitutions block right above it already uses.
- **Split synonyms on `;` as well as `,`.** `austraits.build` declares `(Synonyms, clavate; obclavate)`, which comma-only splitting turned into the single unmatchable synonym `"clavate; obclavate"`.
- **Skip a value that lists itself as its own synonym** (`seed_shape: comma-shaped` does).
- **Tolerate a value with no description** — that aborted with `argument is of length zero`. Latent, not live: 0 occurrences across all three databases today.
- `@param definitions` documented; `man/process_parse_data.Rd` hand-edited rather than roxygenised, since the installed roxygen is 8.0.0 and that migration is separate work on `chore/roxygen-8`.
- `stringr (>= 1.5.0)` in `Depends`, for `str_escape()`.

## Two things for you, not fixed here

1. **13 synonyms in `austraits.build/config/traits.yml` contain a space**, which the code comment said couldn't happen. Two look like APD source defects rather than real synonyms: `sickle-shaped scimitar-shaped` (seed_shape) reads as two synonyms that lost their separator, and `clavate; obclavate` is the semicolon case above. Worth a look upstream in APD, since that's where these are generated.
2. **The `(Synonyms, ...)` convention isn't documented** in the book or the schema — it's currently only discoverable by reading `process_replace_synonyms()`.

## Testing

- `tests/testthat/test-synonyms.R`, 28 assertions: replacement, hyphenated and semicolon-separated synonyms, trait scoping, multi-value cells, whole-word matching (`grapevine` untouched), no-op cases, and the substitution-precedence regression. I checked the precedence test fails under the old order.
- Full suite **937 passing, 0 failures, 0 skips** with `NOT_CRAN=true` (so the `Test_2023_9` snapshot actually runs).
- The nine golden-file examples are **unchanged** — no fixture output moves.
- CI-equivalent check `Status: OK`: `_R_CHECK_CRAN_INCOMING_=false _R_CHECK_FORCE_SUGGESTS_=false NOT_CRAN=true R CMD check --no-manual --as-cran`.

**Not run: the downstream build gate.** This changes trait values in `austraits.build` output, so it wants a rebuild-and-diff there before merge. `ausinvertraits.build` and `AusFizz` declare no synonyms at all, so they're inert.

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -20,7 +20,7 @@ Depends:
     dplyr,
     lubridate,
     readr,
-    stringr,
+    stringr (>= 1.5.0),
     tidyr,
     austraits
 Imports:
--- a/R/process.R
+++ b/R/process.R
@@ -1655,7 +1655,8 @@ process_parse_data <- function(data, dataset_id, metadata, contexts, schema, ide
                 find = stringr::str_split(synonyms_string, ",")[[1]] %>%
                   stringr::str_trim() %>%
                   tolower(),
-                replace = trait_value
+                replace = trait_value,
+                trait_name = trait_name
               )
             } else {
               NULL
--- a/R/process.R
+++ b/R/process.R
@@ -139,7 +139,7 @@ dataset_process <- function(filename_data_raw,
   # Load and clean trait data
   traits <-
     traits %>%
-    process_parse_data(dataset_id, metadata, contexts, schema, identifiers)
+    process_parse_data(dataset_id, metadata, contexts, schema, identifiers, definitions)
 
   # Context ids needed to continue processing
   context_ids <- traits$context_ids
@@ -1408,7 +1408,7 @@ process_add_all_columns <- function(data, vars, add_error_column = TRUE) {
 #' substitutions and unique observation id added
 #' @importFrom dplyr select mutate filter arrange distinct case_when full_join everything any_of bind_cols
 #' @importFrom rlang .data
-process_parse_data <- function(data, dataset_id, metadata, contexts, schema, identifiers) {
+process_parse_data <- function(data, dataset_id, metadata, contexts, schema, identifiers, definitions) {
 
   # Get config data for dataset
   data_is_long_format <- metadata[["dataset"]][["data_is_long_format"]]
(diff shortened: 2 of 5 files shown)


ehwenk (approved):
Thanks for the fix and good to see the overall workflow being merged into develop.

Issue #234, opened 2026-07-28
dataset_test runs on testthat edition 2 while the package is on edition 3


dfalster:
Noticed while adding snapshot coverage of `dataset_test`'s output in #230 (Stage 0 of #225).

The package declares `Config/testthat/edition: 3` and `testthat::edition_get()` confirms it is
active. But the validation layer opts out:

```r
# R/testdata.R:60
local_edition(2)
```

with the comment *"We're using 2nd edition of test that, which has 'context' field"*. That is why
the two `context()` calls at `R/testdata.R:69` and `:935` are still there — one of them marked
`# Keep this` — even though `context()` is deprecated under edition 3 and emits a deprecation
warning on every `dataset_test()` run.

So the edition-3 gap #225 lists as a small fix is not really about the two `context()` calls; they
are a symptom. The question is whether `dataset_test` should move to edition 3.

### What moving would change

- **Reporter output.** Edition 2 groups results by `context()`; edition 3 does not. `dataset_test`
  is the most-taught function in the book, so its console output is user-facing and appears in
  screenshots and tutorials.
- **Expectation semantics.** Edition 3 is stricter: `expect_equal()` uses waldo, and partial
  argument matching and some 2e-era behaviours are gone. Checks that pass today might not.
- **`ListReporter` becomes usable.** Right now `with_reporter(ListReporter, dataset_test_worker(...))`
  errors with *"attempt to use zero-length variable name"* because `context()` is called outside a
  test file. That is why #230 has to capture and filter console text rather than read structured
  results — see the `dataset_test_failures()` helper there.

### Suggested approach

Worth doing, but as its own change with the downstream build gate from #225, not folded into
something else. The snapshot added in #230 pins the current report exactly, so it will show any
wording or ordering change the move causes — which makes this materially safer to attempt now than
it was a week ago.

Related: #213 (turn off backtraces in `dataset_test` output) — #230 strips them before snapshotting
because their line numbers move whenever anything in `R/` is edited, but they are still shown to
users. #129 also touches `dataset_test` ergonomics.

Issue #232, opened 2026-07-28
A `var_in` naming a column that does not exist silently produces an empty identifiers table


dfalster:
Found while closing the output-artefact coverage gap in #230 (Stage 0 of #225).

If a dataset's `identifiers` block names a `var_in` column that is not in `data.csv`, the build
does not complain. It produces an empty identifiers table and carries on.

### Why it is silent

`dataset_process()` passes `identifiers$var_in` to `process_add_all_columns()`
(`R/process.R:127`), which **creates any missing column as all-NA**. The identifiers block then
pivots that column, and `filter(!is.na(identifier_value))` (`R/process.R:206`) drops every row.
No error, no warning, no rows.

### Evidence that this is not hypothetical

`tests/testthat/examples/Test_2023_2` has declared `identifiers: var_in: specimen` since it was
created, and its README has always described the column — but `specimen` was **never committed to
`data.csv` in any revision**. Nobody noticed for the dataset's entire lifetime, because
`identifiers.csv` was not among the tables the tests compared. Its committed 506-row fixture could
not be reproduced from the committed inputs at all.

#230 restores the column so the fixture is meaningful again, and now that `identifiers.csv` is
asserted for all nine examples, a regression to the silent-drop behaviour would show up as
`0 rows != 96 rows`. But that only protects the examples — a real dataset with a typo'd or renamed
`var_in` still loses its identifiers quietly.

### Suggested fix

Error (or at minimum warn) when a declared `var_in` is absent from the data, naming the column and
the `dataset_id`. `dataset_test()` should reject it too.

### Why this needs care

Marked `cross-package` because it changes build behaviour: if any dataset in `austraits.build`,
`ausinvertraits.build` or `AusFizz` currently has a stale `var_in`, this turns a silent no-op into
a hard failure. That is the right outcome, but it should go through the downstream build gate in
#225 rather than land unannounced — and the datasets it catches will need fixing in the same
release.

Same root cause is worth checking for `contexts` and other `var_in`-style declarations.

dfalster:
## Blast radius measured: 0 datasets, but only if the check validates the right thing

I scanned all three downstream repositories locally rather than estimating. This changes the design, so it's worth reading before implementing.

### What's actually out there

| Repo | datasets | declare `identifiers:` | declare a **real** block |
|---|---|---|---|
| `austraits.build` | 411 | 41 | **3** |
| `ausinvertraits.build` | 160 | 0 | 0 |
[...]

PR #262, opened 2026-08-27
Name the dataset, context and column on a bad context var_in

### LINKED ISSUE #247: A contexts var_in naming a missing column fails with an incomprehensible error
Follow-up from the last line of #232, which asked whether `contexts` shares the identifiers root cause. It doesn't — the failure mode is different, and worth its own issue.

## Not silent, but not usable either

Taking `Test_2023_1` and typo'ing one character of a context `var_in` (`sex` → `sexx`), leaving everything else identical:

```
Assigned data `xxx[context_cols[[v]]]` must be compatible with existing data.
✖ Existing data has 1679 rows.
✖ Assigned data has 0 rows.
```

So unlike identifiers before #246, this does **not** silently drop data — it aborts. But the message names **none** of the three things a user needs:

- not the dataset_id
- not the context (`plant sex`)
- not the column that is missing (`sexx`)

`xxx` and `context_cols[[v]]` are internal variable names. Someone hitting this on one dataset out of 411 has no way to tell what to look at, and the message actively misleads — it reads as a data-length problem, not a name typo.

## Mechanism

`process_create_context_ids()` (`R/process.R:790-800`) selects context columns with:

```r
context_cols <- data %>% dplyr::select(dplyr::any_of(tmp$var_in))
```

`any_of()` silently tolerates names that aren't there, so a typo'd `var_in` yields a zero-column selection. The failure surfaces later, as a length mismatch during assignment, far from the cause.

That's the same *shape* as #232 — a permissive selector hiding a bad name — but it manifests as a confusing error rather than data loss, so the fix is diagnostics rather than correctness.

## Suggested fix

Validate context `var_in` values against the data the same way #246 now does for identifiers: after `process_custom_code()` has run, erroring with the dataset_id, the context_property and the missing column name.

```r
missing_var_in <- setdiff(tmp$var_in, names(data))
```

Two things to check that #246 didn't have to worry about:

1. **`var_in` is optional for some context categories.** Some contexts are defined by `values`/`find` rather than by a column, so `tmp$var_in` can legitimately contain `NA` — those must not be treated as missing columns.
2. **Blast radius unmeasured.** #232's turned out to be zero, but that was measured, not assumed. Contexts are used far more widely than identifiers (which only 3 of 601 datasets declare), so this one genuinely needs the downstream scan before implementing — a permissive `any_of()` may be load-bearing for datasets that declare contexts they don't always supply columns for.

## Reproducer

```r
pkgload::load_all("."); setwd("tests/testthat")
schema <- get_schema()
rm2 <- get_schema("config/metadata.yml", "metadata")
defs <- get_schema("config/traits.yml", "traits")
uc <- traits.build:::get_unit_conversions("config/unit_conversions.csv")

md <- readLines("examples/Test_2023_1/metadata.yml")
md[md == "  var_in: sex"] <- "  var_in: sexx"
p <- file.path(tempdir(), "metadata.yml"); writeLines(md, p)

dataset_process("examples/Test_2023_1/data.csv",
                dataset_configure(p, defs), schema, rm2, uc)
```

Unmodified, the same build gives 28 context rows, 15 of them `entity_context`.

dfalster:
A `contexts` `var_in` naming a column that isn't there aborted deep inside `process_create_context_ids()` with ``Assigned data `xxx[context_cols[[v]]]` must be compatible with existing data`` — naming neither the dataset, the context, nor the column, and reading as a data-length problem rather than a typo.

The bad name is now caught in `process_parse_data()` and reported with the dataset_id, the `context_property`, the missing column and the nearest matching column name. Validation runs against the post-`custom_R_code` data plus the fields declared on the metadata `traits` entries, so `var_in: method_context` and `var_in`s created by `custom_R_code` still build. No published trait values move.

Closes #247


dfalster:
Working detail.

## Blast radius: measured, and it is zero

The issue asked for this to be measured rather than assumed, since contexts are used far more widely than identifiers. **348 of 600 datasets across `austraits.build`, `ausinvertraits.build` and `AusFizz` declare contexts. None of them would newly error.**

Getting to that number took two passes, and the first one was wrong in an instructive way. Checking each context `var_in` against the `data.csv` header (post-`custom_R_code`) flags **459** contexts. Of those, 424 are in datasets whose `custom_R_code` creates the column — but the remaining **35 contexts across 32 `austraits.build` datasets** name a column that never exists in the data at all:

```
Blackman_2019   tissue measured             var_in: method_context
Westoby_2004    wood sample type            var_in: method_context
Groves_1986     leaf age                    var_in: method_context_leaf
Kubiak_2009     time to flowering type      var_in: entity_context
Mitchell_2008   sampling season             var_in: temporal_context
... 30 more
```

These build fine today, and a check written against the csv columns would have rejected every one of them. The reason is that `process_create_context_ids()` is called with `out` — the *parsed trait table* — not the raw data. `process_parse_data()` computes `vars_traits <- c(vars, unique(contexts$var_in))` and then, for any `var_in` that is also a field name on the metadata `traits` entries, creates that column from the per-trait values. So `var_in: method_context` is satisfied by

```yaml
traits:
- var_in: WoodD_gcm3
  method_context: wood sample     # <- this is what creates the column
```

The check therefore accepts a `var_in` that names **either** a column of `data.csv` after `custom_R_code`, **or** a field on the `traits` entries. Re-run with that predicate: 0 hits in 348 datasets.

I also built all 32 of those datasets end to end on this branch and on `develop`, and the `contexts` tables are `identical()`.

## Why the check sits in `process_parse_data()` and not in `process_create_context_ids()`

That was my first attempt, and the error message came out actively misleading:

```
- context_property 'plant sex' declares `var_in: sexx`
    (did you mean `instrument used` or `dataset_id` or `method_context`?)
  Available columns: taxon_name, location_name, individual_id, slope position, ...
```

No `sex` anywhere. `df` is built with `select(any_of(c(var_in[i], v, contexts$var_in, identifiers$var_in)))`, so a mistyped `var_in` **drops the very column it meant to name** before the check could see it. Validating earlier, against `data`, is what lets the message offer the intended spelling. (The wild suggestions were `agrep()` matching on substrings; it now uses `utils::adist()` with a length-scaled cap.)

## Before / after

```
Assigned data `xxx[context_cols[[v]]]` must be compatible with existing data.
✖ Existing data has 1679 rows.
✖ Assigned data has 0 rows.
```

```
Dataset Test_2023_1 declares contexts with `var_in` naming a column not present in the data:
  - context_property 'plant sex' declares `var_in: sexx` (did you mean `sex`?)
  A context `var_in` must name either a column of `data.csv` (after `custom_R_code` has run)
  or a field declared on the `traits` entries of `metadata.yml` (e.g. `method_context`).
  Columns in the data: Species, species_abb, ..., sex, slope position, ...
  Fields on the `traits` entries: var_in, unit_in, trait_name, ..., method_context, ...
```

## A second, quieter path fixed alongside

A context that gives no `find`/`value` takes its values from its column. When that column is absent, the `left_join` in `process_format_contexts()` returns zero rows and the context **disappears with no error at all** — the #232 failure mode rather than the #247 one. That now errors too. 0 of 348 datasets are affected (every one of the 35 above declares `values`), but it was one line away and silent.

## Tests

Four new blocks in `test-process.R`: the issue's reproducer, the `var_in: method_context` path (guarded with `expect_false` that it is not a csv column, so the test fails loudly if the fixture stops witnessing it), a `var_in` created by `custom_R_code`, and the no-`values` path.

`NOT_CRAN=true testthat::test_dir()`: **1040 passing, 0 failures, 0 skips** (`NOT_CRAN` matters — the `Test_2023_9` snapshot is `skip_on_cran()`). `R CMD check`: 0 errors, 0 warnings, 0 notes. Golden fixture outputs unchanged.

## Not done here

`dataset_test()` does not check this, so a contributor still learns about a mistyped context `var_in` at build time rather than at test time. The predicate is the same one used here and would drop in cleanly, but it changes `dataset_test()` console output — the most-taught function, and one whose output appears in book screenshots — so it wants its own issue and a maintainer's call, alongside #234.

ehwenk (approved):
This works as intended - I've tested by removing different pieces of context mapped in through the traits section of metadata.yml. You still need to specify the values for such contexts where you map them into the contexts section (unlike for columns in the data.csv file or defined in custom_R_code), because the traits section does not create columns; it just maps to columns. With the changes on this branch, you get a sensible message at least! I think this is the same bug documented in #268 that I documented yesterday.

PR #264, opened 2026-09-04
Parse large locations blocks in chunks (#263)

### LINKED ISSUE #263: metadata.yml locations parse superlinearly (yaml::read_yaml) — 16.8 min for one large georeferenced dataset
## Summary

Loading a dataset's `metadata.yml` via `yaml::read_yaml()` (used in `dataset_configure()`) scales **worse than quadratically** with the number of entries in the `locations:` block. For a dataset with a large number of distinct locations, this dominates build time completely — independent of trait row count.

## Where this showed up

In `austraits.build`, the dataset `AVH_2026` (~366,000 georeferenced herbarium occurrence records) generates one distinct `location_name` per unique lat/long coordinate via its `custom_R_code` (each record has its own coordinate, so effectively no locations are shared/reused). That produces **162,469 distinct entries** in the `locations:` block of `metadata.yml`, a ~488,000-line file.

Measured on that real file:

```r
t0 <- Sys.time()
y <- yaml::read_yaml("data/AVH_2026/metadata.yml")
t1 <- Sys.time()
# Full parse time: 1006.187 s   (~16.8 minutes)
# Number of locations: 162469
```

That's ~17 minutes spent just parsing YAML, before `dataset_process()` does anything else — for comparison, reading/joining the 366K-row trait CSV itself takes seconds (vectorised `readr`/`dplyr`).

## Isolated reproduction

The cost is entirely inside `yaml::read_yaml()`, not traits.build's own downstream code (`process_format_locations()` etc.) — reproduced with synthetic files shaped like a minimal `locations:` block (`location_name: {latitude (deg): ..., longitude (deg): ...}` entries):

| # locations | parse time | ratio per 2× |
|---|---|---|
| 5,000 | 0.27s | — |
| 10,000 | 1.06s | 3.9× |
| 20,000 | 4.55s | 4.3× |
| 40,000 | 23.6s | 5.2× |

Each doubling of location count costs ~4-5× more time (worse than O(n²)), consistent with the ~1000s observed at n=162,469.

Generator used:

```python
def gen(n, out):
    with open(out, 'w') as f:
        f.write("locations:\n")
        for i in range(n):
            lat, lon = -10 - i*0.0001, 130 + i*0.0001
            f.write(f"  at_{lat}_deg_lat_and_{lon}_deg_long:\n")
            f.write(f"    latitude (deg): {lat}\n")
            f.write(f"    longitude (deg): {lon}\n")
```

## Why this will keep recurring

`locations:` in the schema is designed for a small, reusable set of named field sites, but any dataset compiled from per-record georeferenced occurrences (herbarium records, citizen-science observations, GBIF-style compilations) naturally produces one near-unique location per row. As AusTraits absorbs more datasets of this shape, this parsing cost will show up again, and it scales badly enough that it isn't just "a bit slower" — a dataset a few times larger than AVH_2026 would take hours to parse.

Splitting the affected dataset into several smaller sub-datasets works around it on our end for now — since the scaling is superlinear, spreading the same total number of locations across N smaller `metadata.yml` files does much less *total* work, not just the same work in parallel (we measured roughly a 390× reduction in total parse time splitting ~162K locations across 10 files of ~16K each). But this is a workaround, not a fix, and relies on curators noticing the problem.

## Ask

Some combination of:
- A faster parsing path for large `locations:` blocks specifically (alternate YAML parser/backend, or restructuring how very large location sets are represented/loaded so it doesn't go through `yaml::read_yaml()`'s generic nested-list parsing).
- A `dataset_configure()`/validation warning when a dataset's location count crosses some threshold (e.g. a few thousand), pointing curators at this cost and suggesting splitting the dataset.
- Documentation guidance for dataset curators on this failure mode ("don't generate one location per occurrence record; split large per-record-georeferenced compilations into multiple sub-datasets").

Happy to share the full synthetic-benchmark script if useful.

dfalster:
`yaml::read_yaml()` costs roughly O(n^2) in the number of container nodes a document holds rather than in its size, so a dataset generating one location per georeferenced record parses far slower than the trait data it describes: AVH_2026 in austraits.build, with 162,469 locations, took 911s to read.

`read_metadata()` now parses a large `locations:` block in chunks, taking that file to 1.8s and the dataset's build from ~16 min to 46s. Every chunk still goes through the same parser, and any structural surprise falls back to parsing the file whole, so nothing is read differently.

Closes #263 in part; the companion PR adds the csv form.


dfalster:
## Diagnosis

The cost is in **container nodes**, not bytes. Same n, four document shapes:

| shape at n=40,000 | time |
|---|---|
| flat sequence of scalars | 0.02 s |
| sequence of 3-key maps | 6.95 s |
| named map of maps (the `locations:` schema) | 20.3 s |
| flat map, 80k keys | 26.6 s |

Not garbage collection (<1% of elapsed at every size), and writing is linear — `yaml::as.yaml()` on 40k locations is 0.07s. So it is R-object construction on the parse side: a bug in the `yaml` package, not here, and not one worth waiting on.

The lever is that the cost is **per parse call**. Chunking makes it linear.

## Measurements

Real `data/AVH_2026/metadata.yml` (162,469 locations, 17.7 MB, 487,518 lines):

| | parse | dataset build |
|---|---|---|
| `yaml::read_yaml()` | 910.7 s | ~954 s |
| chunked | 1.71 s | 45.7 s |

`process_format_locations()` is linear and was never the problem (28.5 s on those 162,469 locations), but it is now the dominant cost — which is what the companion PR addresses.

## Verification

- `identical()` against `yaml::read_yaml()` on **all 574** `metadata.yml` files in `austraits.build` (414) and `ausinvertraits.build` (160): 0 errors, 0 mismatches.
- The parsed AVH locations match the 910.7 s plain parse.
- Full suite 1072 pass / 0 fail; CI-equivalent `R CMD check` `Status: OK`.

Worth knowing how isolated AVH_2026 is: the **entire rest of austraits.build** parses in **0.6 s combined**, so this one dataset is ~1700× the other 413 put together. The value here is immunising against the next dataset of this shape, not speeding up today's build.

## Two things that bit during development, both now guarded and tested

1. Entry detection must not fire on sequence-valued blocks. `taxonomic_updates:` writes `- find:` at indent 0 with its remaining keys at indent 2, so treating every indent-2 line as an entry start puts chunk boundaries inside entries — this broke ~200 real files on the first attempt.
2. You cannot split the document at column-0 keys generically, because `custom_R_code` holds arbitrary R whose lines are frequently unindented. The block end is anchored on the known top-level block names instead.

## Rejected

- A different YAML backend: there is no maintained alternative parser in R.
- Warning curators above N locations (suggested in the issue): with this in, the threshold where it hurts moves well past anything anyone will hit.

ehwenk (approved):
tested and amazing how much faster this option is

PR #265, opened 2026-09-04
Allow a dataset to hold locations in a csv file (#263)

dfalster:
A dataset compiled from per-record georeferenced occurrences generates one location per record. AVH_2026 in austraits.build has 162,469, making its metadata.yml 487,518 lines and 17.7 MB — neither reviewable nor diffable, and a location table is data rather than configuration.

`locations: locations.csv` now names a file beside metadata.yml holding one row per location. That file goes to 111 lines and the dataset's build to 17s. An empty cell means the property is not recorded, `.na` that it is unknown. Both forms build the same database, verified across every example dataset; nothing needs converting.

Conflicts textually with #264 in `R/utils.R` and `NEWS.md` — see the first comment for the resolution, which is verified green.


dfalster:
## Why a csv rather than only the parser fix in #264

#264 takes the AVH_2026 parse from 911 s to 1.7 s, which leaves `process_format_locations()` — linear, but 28.5 s at 162,469 locations — as the dominant cost. Reading the csv skips both. Measured on the real dataset:

| | metadata.yml | dataset build |
|---|---|---|
| today | 487,518 lines / 17.7 MB | ~954 s |
| with #264 | unchanged | 45.7 s |
| with this | **111 lines / 4.9 KB** (+ 10.6 MB csv) | **17.5 s** |

Same 364,227 trait rows and 324,938 location rows either way. But the file size is the real point: 487,518 lines is not reviewable, greppable or diffable, and a location table is data rather than configuration.

## The one genuine design question: empty vs `.na`

A rectangle cannot say "this location does not record this property", which an inline block says by not listing it. So:

- **empty cell** → property absent for that location, no row, exactly as in the yaml
- **`.na`** → recorded but unknown, keeps its row with a missing value, the same thing `.na` means in the yaml

Read with `na = character()` so neither an empty cell nor the string `NA` is auto-missing — which also means a location genuinely named `NA` survives a write/read round trip (tested).

Without this distinction, a converted dataset whose locations have `.na` coordinates would start failing `dataset_test`'s lat/long presence check. The first draft did exactly that; the round-trip test caught it.

## Verification

The strongest test is `test-locations-csv.R`'s last one: every example dataset with locations is converted to the csv form and built, and **all nine output tables must be equal** to the inline build. Also covered: round trip, empty vs `.na`, the three unusable-file errors, `metadata_add_locations(file =)`, the remake dependency, and `dataset_test` reporting identically on both forms.

Full suite 1128 pass / 0 fail; CI-equivalent `R CMD check` `Status: OK`.

## Integration points that are easy to miss

- **`remake` dependency.** `_config` targets took `metadata.yml` as their only file dependency, so edits to `locations.csv` would not have triggered a rebuild. `build_setup_pipeline()` now emits it.
- **`dataset_test`'s allowed-files whitelist** would have rejected `locations.csv` as a disallowed file.
- **`dataset_test` reads the yaml directly**, not via `read_metadata()` (deliberately — its checks need the yaml as written), so it resolves the csv separately. It also now runs the disallowed-character check over the csv.
- `dataset_test`'s two `!is.null(names(metadata$locations))` guards became `nrow(locations) > 0`, matching what `dataset_process()` already did. These disagreed for a dataset whose whole `locations:` block held a single property — noted in NEWS.

## Merging with #264

Textual conflicts in `R/utils.R` (both insert helpers at the same point; `write_metadata()`'s block list) and `NEWS.md` (adjacent bullets). The semantics compose — `read_metadata()` wants the chunked parser *and* the csv branch. I resolved it in a scratch worktree and ran the merged result: **1173 pass / 0 fail**, `document()` clean, and AVH_2026's inline form still reads in 2.0 s. Merge #264 first and this rebases onto it easily.

## Not done here

The book (`traits.build-book`) documents `locations:` for curators and would want a section on this; it is a separate repo.

dfalster:
## Now also declares the dependency in the `targets` pipeline

`build_targets.whisker` gained the same treatment `build_remake.whisker` already had. Where a dataset's `locations:` names a csv file, the pipeline gets a file target for it and the dataset's `_config` target names that symbol:

```r
tar_target(file_Xxx_locations, "data/Xxx/locations.csv", format = "file"),
tar_target(Xxx_config, {
  file_Xxx_locations
  dataset_configure(file_Xxx_metadata, definitions)
}),
```

The bare reference is load-bearing: `dataset_configure()` reads the csv for itself, and targets works out dependencies from the symbols a command mentions, so without naming it an edit to `locations.csv` would rebuild nothing. Datasets with an inline `locations:` block get the plain one-line `_config` target as before — the section is a whisker inverted section, so nothing is emitted for them.

Verified end to end on a real dataset (`ANBG_2019`, 2,391 locations) converted to the csv form: editing one cell of `locations.csv` and running `tar_make()` rebuilds `file_ANBG_2019_locations` → `ANBG_2019_config` → `ANBG_2019_raw` → `ANBG_2019` → `db`, and the edit lands in the database. Tests assert the file target exists, that it is absent for a dataset without one, and that the `_config` command references it.

## This PR now sits on top of #266

`build_targets.whisker` only exists on the targets branch, so this is rebased onto it. **Merge order: #264, then #266, then this.** The only rebase conflict was `NEWS.md`; `R/setup.R` merged cleanly even though both PRs touch `build_setup_pipeline()`'s whisker values.

## A trap worth recording

While testing this I measured a build in a sandbox with `tar_make()` and got nonsense — locations silently empty, edits not propagating. Cause: `tar_make()` runs the pipeline in a `callr` subprocess that does `library(traits.build)`, which loads the **installed** package, not the `devtools::load_all()` one. Anything testing a change to dataset processing through a targets pipeline has to use `tar_make(callr_function = NULL)` (as the package tests do) or install the branch first, or it is silently testing the released code.

ehwenk (approved):
tested and it works and builds far faster!

PR #266, opened 2026-09-04
Add a targets pipeline that caches like remake (#17)

### LINKED ISSUE #17: [traits.build workflow] Swap out remake for another build pipeline
arising from https://github.com/traitecoevo/austraits.build/issues/354

Austraits.build was using [remake](http://github.com/richfitz/remake), however this package isn't being maintained and will never make it to cran. 

Alternatives include drake, targets, memoise

dfalster:
remake caches, so a curator editing one dataset rebuilds only that dataset — the normal workflow. But remake is unmaintained, off CRAN, and sits in `Remotes`, and furrr parallelises without caching.

`build_setup_pipeline(method = "targets")` writes `_targets.R`. On a 40-dataset compilation: cold 33.5s, one changed `data.csv` 6.9s, nothing changed 0.9s, and 20.0s cold on six crew workers. It caches on target values, so a metadata edit that does not change output stops there. Verified equal to the database `method = "base"` produces.

Closes #17, which was closed in 2024 with targets still outstanding.


dfalster:
## Measured, on a 40-dataset compilation from austraits.build

| | time |
|---|---|
| cold `tar_make()` | 33.5 s |
| cold, 6 crew workers | 20.0 s |
| **one `data.csv` changed** | **6.87 s** |
| nothing changed | 0.87 s |
| (for reference) `method = "base"`, cold | 21.7 s |

Scaled to the full 411-dataset austraits.build, a warm rebuild is roughly independent of dataset count: one dataset (0.36 s) + `bind_databases` (1.98 s) + `saveRDS` (5.62 s) ≈ **8 s against a 154 s cold build**.

## Two behaviours worth knowing

**It caches on target *values*, not timestamps.** Editing `notes:` in a `metadata.yml` rebuilds `file_*_metadata`, `*_config` and `*_raw`, then stops — because `*_raw`'s value is unchanged, the compilation is not rebound or re-saved. remake would have rebuilt everything downstream. Measured at 1.21 s for that case.

**`git_SHA` re-runs every time, by design.** The commit can move without any tracked file changing, so it carries `tar_cue(mode = "always")`. It is cheap, and everything downstream is skipped when the SHA has not moved — which is why the no-op run is 0.87 s rather than paying the 7.6 s tail. This is strictly better than remake's `depends: .git/index`.

## Correctness

All 9 database tables `all.equal()` to the `method = "base"` build on the same 40 datasets. The package test builds the `Test_2022` fixture both ways and compares everything except `build_info` (which records the packages used and legitimately differs, matching how the existing furrr test compares).

Tests use `callr_function = NULL` so the pipeline runs in the testing session against the pkgload-loaded package rather than an installed one, and `skip_if_not_installed("targets")`.

## Cold builds are slower, and that is the trade

33.5 s vs 21.7 s for `base` on the same datasets — targets stores every intermediate. Parallelism recovers most of it (20.0 s on 6 workers), but note the speedup is 1.7x not 6x: per-dataset work averages 0.36 s, so this is serialisation-bound. Do not expect linear scaling.

## Deliberately not done here

- **`remake` is left in place.** Retiring it, and dropping `Remotes: richfitz/remake`, is a separate decision — it is one of the three facts behind the residual `R CMD check` WARNING that #225 tracks, and removing it changes what existing users' repos can do.
- **The `locations.csv` file dependency from #265** needs adding to this template too, exactly as it was added to `build_remake.whisker`. Whichever of the two merges second should carry it.
- **`saveRDS` is 5.62 s of the ~8 s warm floor** at full scale (21 MB, gzip). `compress = FALSE` is 3.54 s but 826 MB; `xz` is 11.7 MB but 28.3 s. Worth a separate decision, and it caps what any caching pipeline can deliver.

dfalster:
## Warm rebuild halved

Of the 6.87 s one-dataset rebuild, only **1.55 s was compute** (`tar_meta` seconds for the 7 rebuilt targets). The rest was `callr` session startup plus serialising the database **three times** — into `austraits_raw`, into `austraits`, and into the export. The base pipeline serialises it once.

Two changes, measured independently on the same 40-dataset sandbox with the same `data.csv` change:

| | cold | warm | warm, no export |
|---|---|---|---|
| as first pushed | 32.9 s | 6.79 s | 5.25 s |
| merged version target | 31.5 s | 5.38 s | 6.10 s |
| qs store format | 26.3 s | 3.98 s | 2.74 s |
| **both (now on the branch)** | **24.6 s** | **3.60 s** | **2.34 s** |

**Merged version target.** `build_add_version()` takes 0.000 s, so splitting it out bought no caching granularity while costing a whole extra copy of the database in the store on every rebuild. Folded into the bind target. The price, accepted deliberately: a commit that changes no data now re-runs `bind_databases()` (~2 s at full scale) where it was previously skipped.

**qs store format.** `format = "qs"` in targets >= 1.12 uses `qs2`, not the older `qs` package. This is targets' *internal* store — the exported `export/data/curr/*.rds` that `austraits` and other consumers read is untouched, so there is nothing for downstream to adapt to. Declared as:

```r
format = if (requireNamespace("qs2", quietly = TRUE)) "qs" else "rds"
```

so it is decided when the pipeline runs, not when it is generated: a curator without qs2 gets a working `rds` pipeline rather than one that fails at `tar_make()`, and installing qs2 later needs no regeneration. `qs2` is in `Suggests`.

## Still correct

All 9 tables `all.equal()` to the `method = "base"` build, `build_info` carries version and SHA. Full suite 1039 pass / 0 fail.

## What is left in the 3.60 s

Roughly 1.2 s is `callr` session startup, which is inherent to `tar_make()`. Dropping the export (`tar_make(names = "database")`) takes it to 2.34 s, so the export is ~1.3 s of what remains — worth knowing for the curator loop, but I have not made the export non-terminal, since that changes what a bare `tar_make()` produces.

dfalster:
## The export is now a step you ask for

`tar_make()` rebuilds the compilation and does not touch `export/data/curr`. `build_export()` writes it.

| | |
|---|---|
| cold `tar_make()` | 22.9 s |
| **one `data.csv` changed** | **2.20 s** |
| `build_export()`, when publishing | 1.28 s |

That is the full arc for this PR: **6.87 s → 2.20 s** on the curator's loop, against 154 s for a cold build of the whole of `austraits.build`.

Why a function rather than a target: `targets` builds every target in the pipeline, and has no per-target "exclude from the default run". Gating on an option or environment variable would have worked but is not something a curator would discover. `build_export()` reads what `tar_make()` already built (`targets::tar_read_raw()`), so there is one source of truth and no risk of exporting something staler than the store.

The generated `_targets.R` says so where the target used to be, and `build_setup_pipeline()` prints it on setup:

```
	-> build compilation using file `_targets.R`, via `targets::tar_make()`
	-> then `build_export()` to write export/data/curr, which `tar_make()` leaves alone
```

## Note for reviewers: this diverges from the other three methods

`base`, `furrr` and `remake` all still write `export/data/curr/<name>.rds` as part of a build; only `targets` now doesn't. That is deliberate — those three have no cache, so every run is a full build and the write is a rounding error — but it does mean `tar_make()` and `source("build.R")` no longer leave the repository in the same state. Worth a conscious yes/no.

Tests cover it: the manifest must not contain `file_database`, and `build_export()` must write a file whose contents match the built compilation.

ehwenk:
I'm not sure if I've run the code properly, but I ran
```
build_setup_pipeline(method = "targets", database_name = "austraits")
austraits <- targets::tar_make("austraits")
```

And ended up with a empty dataset:
```
> nrow(austraits$traits)
NULL
> austraits
NULL
```

ehwenk (approved):
Ok, this works and is fast, although odd the amount of output and that datasets not processed in any sensible order!

But make sure the documentation clearly states how it should be used, because the third step was not obvious to me.

dfalster:
@ehwenk Thanks — the empty result was a real gap in the docs, and chasing it turned up a second trap you would have hit next. Both fixed in cccf01e.

## `austraits <- targets::tar_make("austraits")` returning NULL

Not you running it wrong — that is the direct translation of the `remake` idiom the book still teaches (`austraits <- remake::make("austraits")`), and it cannot work. `tar_make()` builds into the `_targets/` store and returns nothing; you read the result back with `tar_read()`. Your build was fine — it was the assignment that was empty.

So building from `_targets.R` is **three** steps where the other methods are one:

```r
build_setup_pipeline(method = "targets", database_name = "austraits")
targets::tar_make()                        # 1. rebuild
austraits <- targets::tar_read(austraits)  # 2. load it -- tar_make() returns nothing
build_export("austraits")                  # 3. write export/data/curr
```

Verified end to end on a 39-dataset sandbox: 304,801 rows in `austraits$traits`, and the exported `.rds` is `all.equal()` to what step 2 loads.

## The second trap, which you had not reached yet

You set `database_name = "austraits"`. The setup message told you to run `build_export()` — and that would have failed:

```
Error : target database not found
```

because `build_export()` defaults to `"database"`. So step three was not only unobvious, it was also wrong for your pipeline. The messages now interpolate the name you actually used, so all three lines are copy-pasteable, and `build_export()` with a name the pipeline does not have now says why rather than reporting a bare missing target.

The generated `_targets.R` carries the same three steps in a header block, since that is where you look on coming back to a repository. Its header also credited `remake.yml.whisker` — a different template — which is corrected.

## The output, measured against remake

You were right and it is worse than it looks. Same 39 datasets, cold build:

| | lines of output | order |
|---|---|---|
| `remake::make()` | 134 | alphabetical — `ABRS_1981`, `ABRS_2022`, … |
| `targets::tar_make()` | 424 | scheduler order |

Two lines per target instead of one, over 207 targets. On the full 411-dataset `austraits.build` that is roughly 4,400 lines.

**Neither is adjustable, and I would rather say so than pretend.** `targets` collapses only *dynamic branches* into its progress bar; a pipeline of static targets like this one prints per-target lines under every reporter except `"silent"`. And the order is gone by design — `targets` deprecated target priorities in 1.10.1:

> `targets` has moved to a more efficient scheduling algorithm which cannot support priorities. The `priority` argument of `tar_target()` no longer has a reliable effect on execution order.

That is the price of the scheduler that makes the caching fast. What I have done is put the one lever there is at the top of the generated file:

```r
targets::tar_make(reporter = "silent")
```

or `targets::tar_config_set(reporter_make = "silent")` once for the repository. The full three-step flow above runs in **9 lines of output** that way.

Getting real alphabetical progress back would mean restructuring the pipeline around dynamic branching over datasets, which changes the caching granularity that is the whole point of the PR. Worth a separate issue if the noise stays annoying in practice; not worth doing blind.

---

Tests: 921 pass, 0 fail. New tests cover that the setup message and the generated file name all three steps with the database's actual name, and that `build_export()` explains a name mismatch.

**Still outstanding, in the book rather than here:** `adding_data_long.qmd` §Rebuild AusTraits and `tutorial_compilation.qmd` teach `austraits <- remake::make("austraits")`, which is what set the expectation you followed. That is the `traits.build-book` repo — say the word and I will do it as a companion PR.

Branch also brought up to date with `develop`.

PR #269, opened 2026-09-09
Derive context values from `traits` entry literals

### LINKED ISSUE #268: process_format_contexts() crashes with an opaque error when a context has no values: and no matching data.csv column
## Summary

When a `context_property` is populated **only via literal values set on individual `traits:` entries** (no matching `data.csv` column at all — the pattern documented in `file_structure_data_metadata.qmd` as "created ... as elements within specific `traits`"), and its `contexts:` entry omits an explicit `values:` list, `process_format_contexts()` crashes with an opaque internal `rlang`/`dplyr` error rather than a clear, actionable message.

This is adjacent to #247 (same "a permissive selector hides a real problem" shape) but a different failure mode, different function, and — importantly — **not a user mistake**: the `var_in` here is not a typo, it's a name that was never meant to be a `data.csv` column in the first place. Filing separately per the note at the end of #247.

## Background

Per the book, a context's contents can come from either:
1. an actual `data.csv` column (or one created via `custom_R_code`), or
2. a fixed literal value assigned directly on one or more `traits:` entries — e.g.
   ```yaml
   contexts:
   - context_property: wood sample type
     category: method_context
     var_in: method_context
     values:
     - value: stem sample
     - value: branch sample
   traits:
   - var_in: Bark_thickn.br
     ...
     method_context: branch sample
   ```
   Here `method_context` is never a column in `data.csv` — every trait sets it as a literal. (Real example: `Kotowska_2020` in `traitecoevo/austraits.build`.)

For case 1, `values:` can be omitted, and `process_format_contexts()` auto-derives `find`/`value` from the unique values of the matching `data.csv` column (`R/process.R:747-760`). That fallback assumes a matching `data.csv` column always exists. It has no equivalent fallback for case 2 — if you omit `values:` there, it silently tries (and fails) to read a `data.csv` column that was never supposed to exist.

## Reproduction

`metadata.yml` fragment — `var_in: plant_sex` is only ever set as a literal on trait entries, never a `data.csv` column, and `values:` is (incorrectly, but plausibly) omitted:

```yaml
contexts:
- context_property: plant sex
  category: entity_context
  var_in: plant_sex
# no `values:` block, and no "plant_sex" column in data.csv
traits:
- var_in: Corolla_length_male
  trait_name: flower_length
  ...
  plant_sex: male
- var_in: Sepal_female
  trait_name: flower_sepal_length
  ...
  plant_sex: female
```

`dataset_test()` / `build()` crashes deep inside `process_format_contexts()`:

```
Backtrace:
 ...
 └─... %>% dplyr::mutate(find = .data$value) at R/process.R:748:7
 ├─dplyr::left_join(...)
 ├─dplyr::filter(., !is.na(.data$value))
 ├─value
 ├─rlang:::`$.rlang_data_pronoun`(.data, value)
 └─rlang::abort(msg, "rlang_error_data_pronoun_not_found", ...)
```

Real-world case: `Brundrett_2024` in `traitecoevo/austraits.build`, `plant sex` context.

## Mechanism

`process_format_contexts()`, `R/process.R:719-772`. When `values:` is absent, `747-759`:

```r
if (all(!c("find", "value") %in% names(out))) {
  out <- out %>%
    dplyr::select(-any_of(c("value"))) %>%
    dplyr::left_join(
      by = "var_in",
      tibble::tibble(
        var_in = out[["var_in"]][1],
        value = unique(traits[[out$var_in[1]]])   # `traits` here is data.csv
      ) %>% dplyr::filter(!is.na(.data$value))
    ) %>%
    dplyr::mutate(find = .data$value)
}
```

`traits[["plant_sex"]]` doesn't exist → `NULL` → `unique(NULL)` → the constructed tibble collapses to 0 rows/columns → the subsequent `left_join`/`filter(!is.na(.data$value))` has no `value` column to find, and dies with the unrelated-looking `rlang_error_data_pronoun_not_found`. Nothing in the message names the dataset, the context_property, or the missing column.

## Expected behaviour

Either:
1. Also check `metadata$traits` for literal values assigned under a field matching `var_in` when it isn't found as a `data.csv` column, and auto-derive `find`/`value` from those too (description would still need to be added by hand) — removing the need to hand-write `values:` for this case as well; or, at minimum,
2. If `values:` is absent **and** `var_in` names neither a `data.csv` column nor a field actually used on any `traits:` entry, fail immediately with a message naming the dataset_id, the `context_property`, and the `var_in` — e.g. *"context 'plant sex' (var_in: 'plant_sex') has no values: list, and 'plant_sex' is not a column in data.csv — add an explicit values: list."*

## Related

- #247 — same "permissive `any_of()`/missing-name" theme, different function (`process_create_context_ids()`) and different, genuine-typo scenario.
- Once the desired behaviour here is settled, `traits.build-book` needs a documentation update: it currently doesn't state that the "omit `values:`, we'll auto-derive it" shortcut *only* applies when `var_in` names a real `data.csv` column — not when a context is populated purely via literals on `traits:` entries (`file_structure_data_metadata.qmd`, "contexts" section). Happy to file that as a follow-up book issue once this one lands.

dfalster:
A context populated purely by literals set on individual `traits` entries names no `data.csv` column at all. Omitting its `values:` list sent `process_format_contexts()` down the data.csv-only fallback, which read a column that was never meant to exist and aborted with `rlang_error_data_pronoun_not_found` — naming neither the dataset, the `context_property` nor the `var_in`.

Omitting `values:` now derives the values from either route a `var_in` is legitimately satisfied by, matching what `process_parse_data()` already accepts. Descriptions are still hand-written. A `var_in` matching neither is still rejected, and the message now lists both the columns and the `traits` fields rather than reading as though a column were the only option.

Closes #268


dfalster:
Working detail.

**Where the fix sits.** `process_parse_data()` already accepts a context `var_in` satisfied either way — a `data.csv` column, or a field on the `traits` entries — after #262. `process_format_contexts()` runs before it and only had `data.csv` in hand, so it was the one place in the pipeline still treating the column as the only route. It now takes `metadata$traits` as a fourth argument (default `NULL`; the function is not exported, and `responses.build`'s same-named function is its own copy, so no downstream signature breaks).

**Which of the two options in the issue.** Option 1, with option 2 kept as the fallback. Post-#262 the crash was already a clear `stop()`, but it diagnosed a legitimate literal-only context as a misspelled column and blocked a pattern the rest of the build supports — so the clear error alone would not have been enough.

**A value naming a column.** `process_parse_data()` reads a trait-entry field as a column of `data.csv` when the declared value happens to name one, and as a literal otherwise. The derivation here mirrors that (`values_from_trait_entries()`), so a dataset where a literal collides with a column name gets context values that still match the column the build actually produces. Deriving the literal unconditionally would have left those rows with a NA context id and no error — the same silent class of failure #247 was about. Entries with no `trait_name` are filtered out first, since the build drops them.

**Not changed.** A `var_in` that resolves but is all-NA still yields an empty context rather than an error, as before — same outcome for a column and for trait-entry literals, and out of scope here.

**Tests.** Three, in `test-process.R`, all against `Test_2023_1`, which already carries the literal-only pattern (`context_property: branch length`, `var_in: method_context`, values set on the `traits` entries):

- unit: `values:` omitted derives the three branch lengths, `find == value`, `description` NA, and a `trait_name`-less entry's literal does not leak in;
- end-to-end: the fixture rebuilt with that `values:` block deleted produces an identical `contexts` table (bar the hand-written descriptions) and an identical `traits` table;
- the #247 rejection test now also asserts the message names the `traits` fields.

Full suite 1075 passing, 0 failures; golden files unchanged; `lintr` clean on both changed files.

**Follow-up.** The book note at the end of #268 still stands, but says something different now: with this landed, the "omit `values:`" shortcut applies to both routes, so `file_structure_data_metadata.qmd` wants a sentence saying so rather than a caveat. Happy to open that one against `traits.build-book` if useful.

ehwenk (approved):
I've tested removing the values from the context section for contexts added through traits. The trait values do some through in austraits$contexts and no errors! Great!

PR #270, opened 2026-09-09
Convert the dataset report to Quarto and give it real plotting functions

ehwenk:
Splits the report half of `feature/add-dataset-skill` onto its own branch, based on `develop` rather than that branch, so it doesn't drag in the unrelated add-dataset skill.

Run `dataset_report(dataset_id, austraits)` as before — same signature, same output location. What changed underneath:

- `inst/support/report_dataset_2026.Rmd` becomes `inst/support/report_dataset.qmd`, overwriting the old `report_dataset.Rmd` rather than living alongside it.
- `plot_categorical_trait_distribution()`, `plot_trait_distribution_jitter()` and `plot_trait_distribution()` move from `austraits.build/export/plot_trait_values.R` into `R/plot_trait_values.R` as new exported `traits.build` functions — the report (and anyone else) can call them directly, no more sourcing a script by hand.
- `dataset_report()` now renders with `quarto::quarto_render()` instead of `rmarkdown::render()`. Two real consequences of that, not just a rename: Quarto's `execute_params` serialises through YAML, which can't hold an `NA` or an object the size of a compiled database — so `austraits` now goes through a temp `.rds` and the template reads it back by path. And Quarto doesn't bind each param as a bare name the way `rmarkdown` did, so the template now does `dataset_id <- params$dataset_id` explicitly.
- Carries over `new_taxa_trait_combinations()` (the new report calls it) and a small `kable_styling` bugfix — both already on the source branch, needed for the report to build at all.
- CI: `quarto-dev/quarto-actions/setup` added to `R-CMD-check` and `test-coverage`, so the render is actually exercised in CI rather than silently skipped.

**Worth flagging for review:** because Quarto always renders via its own CLI subprocess (which does its own fresh `library(traits.build)`), `dataset_report()` now depends on what's *installed*, not on `devtools::load_all()`. Editing the plotting functions or the template needs a reinstall before the next render reflects it. `R CMD check` installs fresh, so CI isn't affected, but it's a real change to local dev iteration worth being aware of.

Verified: full suite 921/921 passing, including a real end-to-end Quarto render of `Test_2022` (checked for expected content and absence of embedded package errors, not just exit status). `R CMD check`: 0 errors, 0 warnings, 1 NOTE (a nested test-fixture `.git` under `tests/testthat`, pre-existing and unrelated to this change).


R/traits.build-package.R
```diff
@@ -13,6 +13,40 @@ utils::globalVariables(
...
+    "taxon_name",
+    "taxon_rank",
+    "taxonomic_status",
+    "total",
+    "total_counts",
+    "total_per_obs",
+    "trait_name",
+    "ungroup",
+    "value",
+    "value_type"
```
dfalster: Is it Better to add package or .data prefix where these are used?

inst/support/report_dataset.qmd
```diff
@@ -0,0 +1,1176 @@
+---
+title: Report on dataset from `AusTraits` data compilation
```
dfalster, line 2: Claude flagged slightly different workflow for qmd compared to red. Any downsides of this shift? I assume not, as qmd is preferred/modern format?

inst/support/report_dataset.qmd
```diff
@@ -0,0 +1,1176 @@
...
+    theme: yeti
+    toc: true
+    toc-depth: 3
+    embed-resources: true
+params:
+    dataset_id: provide
+    austraits_path: provide
+---
+
+<!-- hack to get indentation on 3rd level of floating TOC; see
```
dfalster: this was needed in Red, is it still needed in qmd?

inst/support/report_dataset.qmd
```diff
@@ -0,0 +1,1176 @@
...
+  $items.each(function(idx) {
+    num_ul = $(this).parentsUntil('#TOC').length;
+    $(this).css({'text-indent': num_ul * 10, 'padding-left': 0});
+  });
+
+});
+</script>
+
+```{r setup, echo=FALSE, message=FALSE, warning=FALSE, results="hide"}
+# knitr defaults
```
dfalster: no longer using Rmd, can any of these be removed?

dfalster (commented):
Nice work, some queries to consider.  just recording them here, next I'll run these through Claude

R/traits.build-package.R
dfalster: Yes — and thanks, this was the wrong tool. Fixed in a349d45.

Two different things were in that list, and `globalVariables()` is wrong for both:

- **Function names** (`extract_dataset`, `extract_trait`, `group_by`, `left_join`, `ungroup`, `row_number`, `n`). Declaring a *function* as a global variable silences a real class of error — a typo'd function name stops being flagged. These resolve at runtime only because `dplyr` and `austraits` are in `Depends`. Now `dplyr::`/`austraits::`.
- **Column names**. Now `.data$` — but only in data-masking verbs.

One wrinkle worth recording, since it changes the answer: **`.data$` is deprecated in tidyselect contexts** as of tidyselect 1.2.0.

```
> select(d, .data$a)
Warning: Use of .data in tidyselect expressions was deprecated in tidyselect 1.2.0.
ℹ Please use "a" instead of .data$a
```

So it is split by verb:

- `mutate()`, `filter()`, `group_by()`, `arrange()`, `distinct()`, `summarise()`, `aes()` → `.data$col` (still recommended for packages)
- `select()`, `pull()`, `across()`, `rename()`, `relocate()` → plain strings or `all_of()`

This is also just the house style — `R/` already has 180 `.data$` and 345 `dplyr::`; `R/plot_trait_values.R` and `new_taxa_trait_combinations()` were the outliers.

`globalVariables()` is back to its original three entries (`.`, `.data`, `..density..`). Verified rather than assumed: `codetools::checkUsagePackage()` reports no undefined globals in the changed functions, and `R CMD check` gives 0 errors / 0 warnings / 1 NOTE (the pre-existing nested `.git`) — no "no visible binding" NOTE.

One thing I did **not** change, flagging it as a possible follow-up: there are 12 pre-existing deprecated `.data$`-in-`pull()` uses elsewhere in `R/` (`setup.R:1608`, `testdata.R`, `process.R`). Out of scope here, happy to do them separately.

inst/support/report_dataset.qmd
dfalster: qmd is the right direction, yes — but there *was* one real downside hiding in this header, which I've now fixed in a349d45.

**`smart: false` silently stopped working.** It was an `html_document` argument; Quarto doesn't recognise it and files it away as unused document metadata. You can see it land in the wrong bucket in the render log — under `metadata`, not under the pandoc options:

```
pandoc
  to: html
  toc: true
  toc-depth: 3
metadata
  smart: false      <- ignored
  theme: yeti
```

Which means smart typography was switching **on** for the first time in this report:

```
input:  A "quoted" phrase, an en--dash, ellipsis...
before: A "quoted" phrase, an en--dash, ellipsis...
after:  A “quoted” phrase, an en–dash, ellipsis…
```

That matters here because the report interpolates metadata text into prose, so a `2"` in a method description or a `--` in a range would have been rewritten. Quarto's equivalent is `from: markdown-smart`, which I've set — verified it restores the straight quotes and that Quarto's own extensions (callouts, cross-refs) still work.

The other consequences of the shift, all in the PR description already: rendering now needs the Quarto CLI, and because Quarto always shells out to it, `dataset_report()` exercises the *installed* package rather than `devtools::load_all()`.

No other downside found — the rendered report is structurally identical (same 14 tables with the same classes, same 10 style blocks, same kableExtra CSS, same 5 widgets and 11 plots).

inst/support/report_dataset.qmd
dfalster: No — removed in a349d45. It can't work under Quarto, and it isn't needed.

**It matches nothing.** The selector is `$('div#TOC li')`. rmarkdown emitted `<div id="TOC">`; Quarto emits:

```html
<nav id="TOC" role="doc-toc" class="toc-active">
```

Checked the real rendered `Test_2022` report — there is no `div#TOC` in it at all, so the selector returns an empty set and the `.each()` is a no-op.

**And it's redundant anyway.** Quarto nests the TOC itself and styles the nesting — the rendered report has 6 `<ul class="collapse">` blocks doing the indentation the hack was hand-rolling.

(Worth noting the hack wasn't failing for the reason you'd first guess. jQuery *is* present in a full report — leaflet drags it in — so it's not a bare `ReferenceError`. It's the selector that's stale. In a report with no widgets there'd be no jQuery either, so it'd fail both ways.)

Rendered before and after: same 27 sections, same TOC nesting, no visual change.

inst/support/report_dataset.qmd
dfalster: Most of it, yes. Done in a349d45 — the setup chunk goes from 5 lines of knitr configuration to 1.

**`knitr::opts_chunk$set(echo = FALSE, cache = FALSE)` → gone.** `echo` belongs in the YAML in Quarto:

```yaml
execute:
  echo: false
```

Which then makes the per-chunk `echo=FALSE` redundant on **all 34 chunks**, so those come off too:

```
- ```{r, results='asis', echo=FALSE}
+ ```{r, results='asis'}
```

`cache = FALSE` was already the default.

**`options(knitr.table.format = "html")` → gone.** This was needed when the output format was ambiguous. Under a Quarto html render `kable()` resolves html on its own — measured in a probe render with the option unset:

```
knitr.table.format option is: NULL
is_html_output: TRUE
table classes: "table table-striped caption-top table-sm small"
```

**`options(dplyr.summarise.inform = FALSE)` → kept.** Still 5 `summarise()` calls in the template.

**The guideline URLs → updated.** They pointed at a 2018 BEES workshop page and `rmarkdown.rstudio.com/lesson-3.html`; now the tidyverse style guide and Quarto's execution-options docs.

**Also `library(knitr)` and `library(kableExtra)` in the next chunk → gone.** Neither is called unqualified anywhere in the template (`util_kable_styling_html()` uses `kableExtra::`). Rather than assume, I rendered with and without and compared: identical table count, identical table classes, kableExtra CSS still embedded, same 10 style blocks. The only byte differences were the random notetaker anchors and htmlwidget ids that differ between *any* two renders — a control render against itself differed by a comparable number of lines.

**And `#' @export` on `as_link()` → gone.** roxygen never sees a template, so it was misleading.

dfalster (approved):

ehwenk:
## Verification found a real regression: taxon-id links break in large real reports

Per a request to verify `dataset_report()` against a real build rather than just the test fixture, I built `Stephens_2024_2` from `austraits.build` (100+ taxa) using this branch and diffed it against the copy already in `austraits.build/export/reports/Stephens_2024_2.html`, built earlier the same day.

**Result: most content matches, but the "Full taxa list" and "Taxonomic updates" tables — the ones with clickable taxon-ID hyperlinks via `as_link()` — come out broken.** Instead of a link, the cell shows literal `[Acacia oxycedrus](https://id.biodiversity.org.au/taxon/apni/51287008)` text, and the table gets pandoc's own row styling instead of the bootstrap classes `my_kable_styling()` sets.

### How I found and traced it

1. **First pass, comparing full rendered text.** Extracted visible body text from old vs. new HTML (`rvest::html_text2`, stripped `<script>`/`<style>`) and diffed word-by-word rather than line-by-line, to avoid false positives from re-wrapped paragraphs. Most differences were harmless engine artifacts (curly vs. straight quotes, a TOC listing rvest picks up differently). One block of differences was not: every taxon name in the "Full taxa list" table showed the raw markdown/HTML source instead of a rendered link.

2. **Root cause #1 (real, fixed): `as_link()` was still emitting markdown syntax (`[text](url)`) into a `kableExtra`-rendered HTML table cell.** rmarkdown's pandoc reparses markdown found inside a raw HTML block (`markdown_in_html_blocks`); Quarto's does not. Switched both call sites to `as_link(..., type = "html")` (a real `<a>` tag).

3. **Root cause #2 (real, fixed): `escape = FALSE` alone didn't survive.** Isolated test: `util_kable_styling_html(x, escape = FALSE)` still HTML-escaped a literal `<a>` tag into `&lt;a href=...&gt;`, because `kable_styling()` falls back to its own default escaping unless the table it receives is *already* HTML-formatted — and under Quarto, `options(knitr.table.format = "html")` didn't reliably make it so. Fix: pass `format = "html"` explicitly alongside `escape = FALSE`.

4. **Root cause #3 (real, fixed): the chunk needed to emit its output as guaranteed raw HTML.** Comparing against the ~16 *other* tables in the document that render correctly, the working ones all end in `%>% writeLines()`; mine returned the styled table as a bare final expression. Added `writeLines()`, and — matching what those other tables get automatically from knitr — wrapped the whole thing in an explicit `` ```{=html} `` / `` ``` `` fence.

5. **Still broken after all three fixes — bisected further.** To find where the corruption actually happens, I:
   - Dumped the *intermediate* markdown Quarto hands to pandoc (`quarto_render(..., debug = TRUE)` keeps it as `<file>.html.md`) and checked it directly: **100% correct** — proper `<a href>` tag, proper bootstrap classes, properly fenced.
   - Extracted just that fenced block (with the real YAML header, real `toc`/`theme` options) into a standalone minimal `.qmd`/`.md` and rendered it alone: **renders correctly** — link intact (Quarto does rewrite the table's classes for its own styling, but the link itself survives).
   - Took the *entire real prefix* of the document (everything before the table, ~845KB) plus the table, rendered as a standalone `.md`: **still correct**.
   - Rendered the *complete* real intermediate `.md` file (using `quarto render <file>.md` directly, bypassing re-execution): **still correct.**
   - Only going through the full `.qmd` pipeline on the complete, real, ~1200-line document — R execution + knit + Quarto's own pandoc invocation — reproduces the break.
   - Checked for the obvious explanation (an unbalanced/mismatched code fence earlier in the document corrupting parser state for everything downstream): wrote a fence-balance tracker over the whole intermediate markdown; found no imbalance up to the point our table starts.

So: the generated content is correct, isolated reproductions of the exact same content all render correctly, and I have not found what's different about the full real document. This looks like a genuine, scale-or-complexity-dependent difference in how Quarto invokes pandoc for a `.qmd`-executed document vs. a plain `.md` file, rather than anything reachable by editing the template's content. I re-tested after rebasing onto `a349d45` (dropping `options(knitr.table.format = "html")` etc.) — still reproduces.

### What this means for the PR

Fixes 2–4 above are genuine improvements and are pushed (`2072270`). But **this PR has not been verified equivalent to the old report for any dataset with a real "Full taxa list" or "Taxonomic updates" table** — which is effectively every dataset, since every one has taxa. I'd hold off merging until this is resolved or the taxon-link tables are handled a different way (e.g. post-processing the rendered HTML string after `quarto::quarto_render()` completes, entirely outside pandoc's control, rather than relying on the raw-HTML-block surviving Quarto's own table pass).

ehwenk:
## Resolved: render in two Quarto passes

Found it. The bisection in the previous comment established that the *exact same* knitted markdown content rendered correctly whenever it went through `quarto render <file>.md` directly, and only broke when it went through the full `.qmd` pipeline (`quarto_render()` on the `.qmd` itself). That's the whole fix: Quarto's pandoc invocation for a `.qmd` input reprocesses the taxon-list tables' raw `` ```{=html} `` blocks differently than it does for a plain `.md` file, discarding the `<a>` tags. Rendering the same knitted content a second time as markdown does not have this problem.

`dataset_report_worker()` now renders in two passes:
1. `quarto::quarto_render(input_qmd, ..., debug = TRUE)` — knits and produces its own (broken) html, which is discarded; `debug = TRUE` is what leaves the knitted `.md` in place afterward.
2. `quarto::quarto_render(knitted_md, ...)` — renders *that* markdown to the final html.

Cleans up the extra intermediate (`.html.md` and its `_files` directory) alongside the `.qmd` afterward, same as before.

**Verified against the same real case that exposed this** (`austraits.build`'s `Stephens_2024_2`, 100+ taxa) via the actual `dataset_report()` call, not a synthetic test:
- `](https` literal brackets: 34 → 0
- Bootstrap `table-striped` tables: 16/16 (matches the pre-existing report)
- Full word-level text diff against the pre-existing `export/reports/Stephens_2024_2.html`: nothing left beyond known engine cosmetics (curly-vs-straight quotes from the `from: markdown-smart` change, a TOC-listing text-extraction artifact `rvest` picks up differently, `sessionInfo()` naturally differing between machines/times)

Full suite 921/921. `R CMD check`: 0 errors, 0 warnings, 1 pre-existing NOTE (the same unrelated nested test-fixture `.git`, not this PR's doing).

I'd consider this ready to merge from the report/plotting side now — happy to take another look if anything else turns up in review.

